PR #6, opened 2025-08-20
fix(subtask): Prevent subtask from reading custom rules

Sma1lboy:
## Summary
- Prevents sub-tasks from accessing and reading custom rules defined in the project. This enhances security and encapsulation by ensuring sub-tasks operate only with the permissions they are explicitly granted.

## Test plan
- Manually verified that sub-tasks no longer have access to custom rules.


packages/tools/src/new-task.ts
```diff
@@ -14,6 +14,8 @@ const toolDef = {
   description:
     `Create a task that can be executed autonomously by a runner in the same environment as the current task. When you are searching for a keyword or file and are not confident that you will find the right match in the first few tries, use the Agent tool to perform the search for you.
 
+Note: The sub-agent operates independently and does not inherit your custom instructions or rules. Keep the task description focused on the specific objective without referencing personal preferences or custom rules.
+
```
wsxiaoys:
```suggestion
```

packages/vscode-webui/src/features/chat/lib/use-live-chat-kit-getters.ts
```diff
@@ -31,13 +33,23 @@ export function useLiveChatKitGetters({
...
+      const envToReturn = isSubTask
+        ? {
+            ...environment,
+            info: {
+              ...environment.info,
+              customRules: undefined,
+            },
+          }
+        : environment;
+
```
wsxiaoys: pass isSubTask to readEnvironment, and control customRule behavior in readEnvironment instead

wsxiaoys (approved):

PR #68, opened 2025-08-25
feat(cli): enhance model configuration and CLI options

wsxiaoys:
## Summary
- Update CLI version to 0.5.0-dev
- Restructure model configuration with new modelType parameter (pochi/openai)
- Add model-specific options with proper defaults
- Improve option parsing and validation
- Update user agent string to use "Pochi" instead of "PochiCli"
- Add default values for maxRounds (24) and maxRetries (3)
- Enhance API client authentication handling
- Improve error handling for option validation

## Test plan
- [x] All existing tests pass
- [x] CLI functionality verified
- [x] Model configuration options work correctly
- [x] Authentication handling works as expected

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -80,17 +80,17 @@ const program = new Command()
   )
   .option(
     "--model-api-key <modelApiKey>",
-    "The API key to use for authentication. Only meant to be set for `openai` models.",
+    "The API key to use for authentication. Used for all non-pochi models (e.g., `openai`).",
   )
   .option(
     "--model-max-output-tokens <number>",
-    "The maximum number of output tokens to use. Only meant to be set for `openai` models.",
+    "The maximum number of output tokens to use. Used for all non-pochi models (e.g., `openai`).",
     parsePositiveInt,
     4096,
   )
   .option(
     "--model-context-window <number>",
-    "The maximum context window size in tokens. Only meant to be set for `openai` models.",
+    "The maximum context window size in tokens. Used for all non-pochi models (e.g., `openai`).",
     parsePositiveInt,
     100_000, // 100K
   )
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli.ts
```diff
@@ -172,7 +172,7 @@ async function parseTaskInput(options: ProgramOpts, program: Program) {
     }
   }
 
-  if (!uid && !prompt) {
+  if (!prompt) {
```
zwpaper, line 175: should update the error output in this case

packages/cli/src/cli.ts
```diff
@@ -22,84 +22,84 @@ logger.debug(`pochi v${packageJson.version}`);
...
+    "pochi",
+  )
+  .option(
+    "--model-base-url <baseURL>",
+    "The base URL to use for the model API.",
+    prodServerUrl,
+  )
+  .option(
+    "--model-api-key <modelApiKey>",
+    "The API key to use for authentication. Only meant to be set for `openai` models.",
```
zwpaper: no only for `openai`?
i saw both `ai-gateway` and `google-vertex-tuning` used this option.

packages/cli/src/cli.ts
```diff
@@ -22,84 +22,84 @@ logger.debug(`pochi v${packageJson.version}`);
...
+    "--model-max-output-tokens <number>",
+    "The maximum number of output tokens to use. Only meant to be set for `openai` models.",
     parsePositiveInt,
+    4096,
   )
   .option(
-    "--context-window <number>",
-    "The context window limit for BYOK model.",
+    "--model-context-window <number>",
+    "The maximum context window size in tokens. Only meant to be set for `openai` models.",
```
zwpaper: ditto

packages/cli/src/cli.ts
```diff
@@ -22,84 +22,84 @@ logger.debug(`pochi v${packageJson.version}`);
 
 const prodServerUrl = "https://app.getpochi.com";
 
-const userAgent = `PochiCli/${packageJson.version} ${`Node/${process.version}`} (${process.platform}; ${process.arch})`;
+const userAgent = `Pochi/${packageJson.version} ${`Node/${process.version}`} (${process.platform}; ${process.arch})`;
```
zwpaper: should we distinguish between cli and vscode?

PR #146, opened 2025-09-01
feat: Add barebones upgrade functionality

### LINKED ISSUE #72: feat(cli): support self upgrade with 'pochi upgrade'
---
name: Feature request
about: Suggest an idea for this project
title: 'feat(cli): support self upgrade with `pochi upgrade`'
labels: enhancement
assignees: ''

---

It would be great to have a self-upgrade mechanism for the Pochi CLI. A command like `pochi upgrade` could check for a new version on GitHub Releases and, if a newer version is available, download and replace the current binary.

This would simplify the update process for users, ensuring they can easily stay on the latest version.

**Implementation Details:**

*   The command should be added to `packages/cli/src/cli.ts` (https://github.com/TabbyML/pochi/blob/main/packages/cli/src/cli.ts).
*   It will need to fetch release information from the GitHub API for the `TabbyML/pochi` repository.
*   It should compare the latest release tag with the current version.
*   If a new version exists, it should download the appropriate binary for the user's OS and architecture, and replace the existing executable.

**Additional context**

This feature will improve the user experience and make it easier to keep the CLI tool up-to-date.

---

DESU-CLUB:
Closes #72 

Implemented a barebones upgrade extension for Pochi, using pochi upgrade.

#### Building from source (github)
1. Run `bun run build`
2. Run `./dist/pochi upgrade`
3. Follow instructions to alias pochi to `~/.pochi/bin`
4. Check with pochi -V (Should be updated to latest version

#### Building from bun run
1. Run `bun run pochi upgrade`
2. Follow instructions to alias pochi to `~/.pochi/bin`
3. Check with pochi -V

### How does it work
- Implemented upgrade in cli.ts, to call the src/lib/upgrade.ts
- src/lib/upgrade.ts does a fetch to tabbyml latest github release, filter by CLI (Code is aware of VSCode releases, and will filter out)
- Check platform of user, and download correct version (zip for Windows, MacOS and Linux uses tar.gz)
- Download to ~/.pochi
- Alias from there and you should have most recent version

### TODO
- [x] Sync with TabbyML team on future proposed direction and code smell check
- [x] Update documentation
- [x] Update Pochi cli help to reflect upgrade
- [x] Investigate better ways to handle downloading + installation of new Pochi binaries
- [x] Test on Windows and MacOS
- [x] Add feature to unlink existing pochi binaries, and replace with new pochi binary from upgrade

### Known Bugs
- At this point of time, upgrade.ts is not integrated into Pochi 0.5.7, therefore there might be bugs on trying to replicate upgrade behaviour on newest version of Pochi.

### DIFF
--- a/packages/vscode-webui/src/components/prompt-form/auto-completion/extension.ts
+++ b/packages/vscode-webui/src/components/prompt-form/auto-completion/extension.ts
@@ -236,6 +236,10 @@ export const AutoCompleteExtension = Extension.create<
               return;
             }
 
+            if (isQueryExactMatch(props)) {
+              return;
+            }
+
             storage.component = new ReactRenderer(AutoCompleteMentionList, {
               props: { ...props, fetchItems },
               editor: props.editor,
@@ -300,7 +304,11 @@ export const AutoCompleteExtension = Extension.create<
               const suggestionActive = suggestionTriggerPluginKey.getState(
                 props.editor.state,
               );
-              if (!props.items?.length || !suggestionActive) {
+              if (
+                !props.items?.length ||
+                !suggestionActive ||
+                isQueryExactMatch(props)
+              ) {
                 destroyMention();
                 return;
               }
@@ -341,4 +349,16 @@ function isMentionExtensionActive(state: EditorState) {
   const fileMentionState = fileMentionPluginKey.getState(state);
   const workflowMentionState = workflowMentionPluginKey.getState(state);
   return fileMentionState?.active || workflowMentionState?.active;
+}
+
+/**
+ * Checks if the query exactly matches the only candidate.
+ * @param props The suggestion props.
+ * @returns `true` if there is exactly one candidate and its value matches the query, otherwise `false`.
+ */
+function isQueryExactMatch(
+  props: SuggestionProps<AutoCompleteSuggestionItem>,
+): boolean {
+  const { query, items: candidates } = props;
+  return !!query && candidates.length === 1 && candidates[0].value === query;
 }
--- a/packages/vscode-webui/src/components/prompt-form/auto-completion/mention-list.tsx
+++ b/packages/vscode-webui/src/components/prompt-form/auto-completion/mention-list.tsx
@@ -61,7 +61,7 @@ export const AutoCompleteMentionList = forwardRef<
   useImperativeHandle(ref, () => keyboardNavigation);
 
   return (
-    <div className="relative flex w-[80vw] flex-col overflow-hidden py-1 sm:w-[600px]">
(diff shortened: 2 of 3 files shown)


packages/cli/src/lib/binary-installer.ts
```diff
@@ -0,0 +1,280 @@
...
+      } catch (error) {
+        // Fallback to execSync if symlinkSync fails
+        execSync(`ln -s "${versionedBinaryName}" "${latestBinaryPath}"`, {
+          cwd: binDir,
+        });
+      }
+    }
+
+    // Download and copy the install script
+    await downloadAndCopyInstallScript(binDir);
```
wsxiaoys: seems pochi is already downloaded - why there's a need to download install.sh?

packages/cli/src/lib/release-fetcher.ts
wsxiaoys, line 1: consider create a `lib/install` package to packaging new packages.

wsxiaoys (changes requested):

packages/cli/src/lib/binary-installer.ts
DESU-CLUB: Removed the install.sh functionality

packages/cli/src/lib/install/version-utils.ts
```diff
@@ -0,0 +1,37 @@
+// Compare two versions and return true if the latest is newer
+export function isNewerVersion(latest: string, current: string): boolean {
```
zwpaper: we should support the version with `-dev`, `-rc` suffix,
the current impl would be break when having suffix

```
Checking for updates...
Current version: 0.5.10-dev
Latest version: 0.5.9
A new version (0.5.9) is available!
Downloading pochi-mac-arm64.tar.gz...
Installing to: /Users/kw/.pochi/bin
```

DESU-CLUB: Got it let me add that as well!

DESU-CLUB: Can I check for this if I have 0.5.10-dev, 0.5.10-rc, 0.5.10, in which order does it take precedence?

Right now the comparison is done via the numbers, so if 0.5.10-rc is the newest version and 0.5.10 is our current version no upgrade will be done

DESU-CLUB: Resolved the emojis, and a barebones implementation of handling suffixes (-dev, -rc) has been done. For now, intended behaviour is not to update when major and minor versioning numbers are same, as from my observation these are not found on the github releases page

PR #250, opened 2025-09-09
feat(agent): add agent parser and loader

zhanba:
## Screenshot
image

## Summary
This PR introduces the following features:
- An agent parser to handle agent configurations.
- An agent loader to load agents in the CLI.

## Test plan
- Unit tests for the agent parser and loader have been added.

### DIFF
--- a/packages/cli/src/types.ts
+++ b/packages/cli/src/types.ts
@@ -1,3 +1,4 @@
+import type { CustomAgent } from "@getpochi/tools";
 import type { TaskRunner } from "./task-runner";
 
 export interface ToolCallOptions {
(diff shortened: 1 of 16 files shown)


packages/cli/src/types.ts
```diff
@@ -14,8 +15,16 @@ export interface ToolCallOptions {
...
+   */
+  customAgents?: CustomAgent[];
+
   /**
    * Function to create a sub-task runner (optional, used by newTask tool)
    */
-  createSubTaskRunner?: (taskId: string) => TaskRunner;
+  createSubTaskRunner?: (
+    taskId: string,
+    customAgent?: CustomAgent,
```
wsxiaoys, line 28: we don't allow `newTask` for subtask, thus no need to pass this?

PR #260, opened 2025-09-09
chore(action): move reaction behavior into first step of action

Sma1lboy:
This pull request introduces improvements to the GitHub Action workflow for the Pochi CLI by moving the creation of the initial progress comment and the "eyes" reaction from the TypeScript implementation to the shell script steps in the workflow YAML. This change streamlines the workflow setup and ensures that environment variables for these resources are available to subsequent steps and the CLI itself.

**Workflow enhancements:**

* Added a new shell script step in `action.yml` to create the initial "eyes" reaction and progress comment via GitHub API calls, exporting their IDs as environment variables for use in later steps.
* Updated environment variables in the workflow to include `PROGRESS_COMMENT_ID` and `EYES_REACTION_ID`, making them available to the Pochi CLI execution step.

**Codebase adjustments:**

* Modified `run-pochi.ts` to use the pre-created comment and reaction IDs from environment variables if available, falling back to creating them in code only if not set. This avoids duplicate API calls and leverages the workflow setup.

### DIFF
--- a/packages/github-action/action.yml
+++ b/packages/github-action/action.yml
@@ -19,11 +19,6 @@ runs:
         COMMENT_ID=$(echo '${{ github.event.comment.id }}')
         ISSUE_NUMBER=$(echo '${{ github.event.issue.number }}')
         
-        # Debug output
-        echo "Debug: COMMENT_ID=$COMMENT_ID"
-        echo "Debug: ISSUE_NUMBER=$ISSUE_NUMBER"
-        echo "Debug: REPOSITORY=${{ github.repository }}"
-        
         # Add eyes reaction to indicate starting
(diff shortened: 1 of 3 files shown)


Sma1lboy:
/pochi review this pr to give any suggestion if you have

packages/github-action/action.yml
```diff
@@ -12,6 +12,39 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Add initial reaction and create progress comment
```
wsxiaoys: would it easier to wrap these logic in typescript, or use `gh` command line, rather than invoking api by your self?

Issue #225, opened 2025-09-07
 cli render get error when search tool return 0 match


ZHallen122:
**Describe the bug**
When the search tool returns 0 matches in cli. Then the Cli render would show an error. It won't stop cli.
To recurrence the bug use this in cli: "pochi -p "Use search tool and cause it to return zero match""

<img width="1313" height="383" alt="Image" src="https://github.com/user-attachments/assets/aafde9b3-7090-471b-a5ca-480670e577e5" />

**Information about your version**
This issue was observed in the Pochi CLI running from the latest `main` branch 9/7/2025 version.

🤖 Generated with [Pochi](https://getpochi.com)

wsxiaoys:
<img width="770" height="272" alt="Image" src="https://github.com/user-attachments/assets/884610e3-b516-4fb7-bd7a-d1e44aba3c4a" />

<img width="1367" height="272" alt="Image" src="https://github.com/user-attachments/assets/87bea065-fe1b-455d-824e-dce1bd21f587" />

PR #263, opened 2025-09-09
fix(ripgrep): improve error logging for rg command execution

Sma1lboy:
Relate https://github.com/TabbyML/pochi/issues/225

image

This pull request makes a minor adjustment to error logging in the `searchFilesWithRipgrep` function. The change ensures that the error is only logged if it does not satisfy the `ExecError` type, improving the clarity and accuracy of error reporting.


wsxiaoys:
/pochi is there document need to be update with this?

PR #283, opened 2025-09-11
fix(cli): check for existing share link before creating a new one

Sma1lboy:
This pull request refactors the task sharing command in the CLI to simplify its logic and improve clarity. The command now retrieves an existing share URL for a task instead of creating a new share link, and unnecessary code for API calls and error handling has been removed.

**Command and user experience improvements:**

* Renamed the command from `share` to `get-share-url`, updating the description and argument to reflect that it retrieves an existing share URL rather than creating a new share link.
* Improved user feedback by changing output messages to reference "Share URL" and providing clearer error messaging when no share URL is found.

**Codebase simplification:**

* Removed the `createShareLink` function and all related API client logic, so the command now only queries the local store for the share ID.
* Eliminated error handling and asynchronous API calls, further streamlining the code and reducing complexity.

### DIFF
--- a/packages/cli/src/task/share.ts
+++ b/packages/cli/src/task/share.ts
@@ -1,91 +1,30 @@
 import type { Command } from "@commander-js/extra-typings";
-import { type Message, catalog } from "@getpochi/livekit";
+import { catalog } from "@getpochi/livekit";
 import chalk from "chalk";
-import { createApiClient } from "../lib/api-client";
 import { createStore } from "../livekit/store";
 
 export function registerTaskShareCommand(taskCommand: Command) {
-  // pochi task share <id> - Create share link for a task ID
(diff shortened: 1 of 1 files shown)


packages/cli/src/task/share.ts
```diff
@@ -45,6 +45,15 @@ async function createShareLink(
...
 ): Promise<string | null> {
   try {
+    // First check if shareId already exists locally
+    const taskData = store.query(catalog.queries.makeTaskQuery(taskId));
+
+    if (taskData?.shareId) {
+      console.log(chalk.gray("Share link already exists locally."));
+      return taskData.shareId;
+    }
+
```
wsxiaoys: you can simply remove part below, and maybe rename command to `get-share-url` instead of `share`

PR #293, opened 2025-09-14
feat(livekit-cf): implement task persistence to Pochi API

wsxiaoys:
## Summary
- Add moment.js dependency for time handling in Cloudflare Workers
- Implement task persistence logic in LiveStoreClientDO to persist tasks to Pochi API
- Add POCHI_API_KEY to Cloudflare environment variables
- Remove old background job persistence in favor of DO-based approach
- Add endpoint to activate store
- Export events from livestore schema
- Fix type errors in livekit package after refactoring

This change moves task persistence from the client-side background jobs to the Cloudflare Durable Objects, ensuring better reliability and consistency.

### DIFF
--- a/packages/livekit-cf/src/client/do.ts
+++ b/packages/livekit-cf/src/client/do.ts
@@ -1,12 +1,16 @@
 import { DurableObject } from "cloudflare:workers";
 import type { Env } from "@/types";
-import { catalog } from "@getpochi/livekit";
+import type { PochiApi, PochiApiClient } from "@getpochi/common/pochi-api";
+import { decodeStoreId } from "@getpochi/common/store-id-utils";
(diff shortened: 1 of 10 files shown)


packages/livekit-cf/src/client/do.ts
```diff
@@ -94,4 +99,79 @@ export class LiveStoreClientDO
...
+                  mainBranch: "",
+                  currentBranch: task.git.branch,
+                  recentCommits: [],
+                }
+              : undefined,
+          },
+          todos: task.todos as DeepWriteable<typeof task.todos>,
+        },
+      },
+    });
```
wsxiaoys, line 148: @zwpaper fyi (pochi task 273)

PR #294, opened 2025-09-14
feat(cli): add upgrade information on startup

### LINKED ISSUE #277: feat(cli): display upgrade information on startup
**Is your feature request related to a problem? Please describe.**
Currently, the CLI only checks for updates when running `pochi upgrade`

**Describe the solution you'd like**
We should add a version check to the main CLI entry point that runs before any other command logic. If a new version is available, a notification should be displayed in the console.
unction.
3.  In , call  at the beginning of the main  block.

**Additional context**
Here is a screenshot of the desired behavior from (wrangler):

Image

Relevant code:
*   [](https://github.com/TabbyML/pochi/blob/upgrade-live-store-add-sync/packages/cli/src/upgrade/cmd.ts)
*   [](https://github.com/TabbyML/pochi/blob/upgrade-live-store-add-sync/packages/cli/src/cli.ts)

DESU-CLUB:
Closes #277 

### Summary
- Adds a startup routine that prints the current version and whether an update is available
- Refactored internal logic for cmd.ts 
  - `checkForUpdates` (packages/cli/upgrade/version-check.ts) and `registerUpgradeCommand`  (packages/cli/upgrade/cmd.ts) now internally use `returnVersionInfo`, to de-duplicate potentially redundant code
- Added program preAction hook that runs version-check before any command

#### Building from source (Github)
- Fetch this pull request
- Run `bun install` in the root folder
- Run `bun run build` in packages/cli
- Run `./dist/pochi -<command>`: The upgrade information should be shown on display

#### Running using bun
- Fetch this pull request
- Run `bun install` in the root folder
- Run `bun run pochi -<command>`: The upgrade information should be shown

### Known Issues
- The display feature does not show during pochi -V and pochi -h. This is as CommanderJS do not treat the help and version actions as commands and exits early.

### TODO
- [ ] Check with Tabby team if the upgrade display should show when pochi -V (show current version) and pochi -h (Show help) [Current implementation does not show when these commands are called]
- [ ] Check with Tabby team if the console.log in upgrade command stating new version is redundant
- [x] Test on Windows and MacOS machines
- [ ] ~~Update documentation with note that now upgrade is displayed on startup~~ (Doesn't seem like it belongs in docs, which talks about core features rather than UI changes) 

### Display Preview 

image

*Display Preview with `prompt` command*

image

*Display Preview with `upgrade` command*

### DIFF
--- a/packages/cli/src/upgrade/cmd.ts
+++ b/packages/cli/src/upgrade/cmd.ts
@@ -13,8 +13,12 @@ export function registerUpgradeCommand(program: Command) {
       console.log("Checking for updates...");
 
       try {
-        const { updateAvailable, latestVersion, latestRelease, currentVersion } =
-          await returnVersionInfo({ timeoutMs: null });
+        const {
+          updateAvailable,
+          latestVersion,
+          latestRelease,
+          currentVersion,
+        } = await returnVersionInfo({ timeoutMs: null });
         console.log(`Current version: ${currentVersion}`);
 
         if (updateAvailable) {
--- a/packages/cli/src/upgrade/version-check.ts
+++ b/packages/cli/src/upgrade/version-check.ts
@@ -55,5 +55,5 @@ export async function checkForUpdates() {
     const columns = process.stdout.columns || 80;
     const width = Math.max(Math.min(columns, 100), 20);
     console.log(chalk.yellow("─".repeat(width)));
-  } 
+  }
 }


packages/cli/src/upgrade/version-check.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface VersionCheckResult {
+  updateAvailable: boolean;
+  currentVersion: string;
+  latestVersion: string;
+  latestRelease: GitHubRelease;
+}
+
+export async function returnVersionInfo(): Promise<VersionCheckResult> {
+  const latestRelease = await fetchLatestCliRelease();
```
wsxiaoys: consider add a timeout (e.g 200ms) to avoid blocking the command for too long

DESU-CLUB: Have added the timeout in the latest push! After experimenting on my machine, I set the timeout to 300ms (It was not able to fetch the release details within 200ms), however we might want to investigate more if we should extend the timeout to be longer in the scenario of poor wifi connectivity

wsxiaoys: that'll make a good case that the short timeout shall only apply for auto version check, but not for `pochi upgrade` itself - so user have a choice to explicit upgrade without being restricted by timeout

DESU-CLUB: Got it! Have added this in the latest iteration, such that only version check uses the 300ms timeout, while pochi upgrade awaits without being restricted by timeout!

packages/cli/src/upgrade/version-check.ts
```diff
@@ -0,0 +1,61 @@
...
+  updateAvailable: boolean;
+  currentVersion: string;
+  latestVersion: string;
+  latestRelease: GitHubRelease;
+}
+
+export async function returnVersionInfo(options?: {
+  timeoutMs?: number | null;
+}): Promise<VersionCheckResult> {
+  const { timeoutMs } = options ?? {}; //Optional timeout, used for version check
```
zwpaper:
```suggestion
  const { timeoutMs } = options ?? {}; // Optional timeout, used for version check on startup
```

packages/cli/src/upgrade/version-check.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const latestVersion = extractVersionFromTag(latestRelease.tag_name);
+  const currentVersion = packageJson.version;
+
+  return {
+    updateAvailable: isNewerVersion(latestVersion, currentVersion),
+    currentVersion,
+    latestVersion,
+    latestRelease,
+  };
```
zwpaper:
```suggestion
  return {
    updateAvailable: isNewerVersion(latestVersion, currentVersion),
    currentVersion,
    latestVersion: extractVersionFromTag(latestRelease.tag_name),
    latestRelease: packageJson.version,
  };
```

packages/cli/src/upgrade/version-check.ts
```diff
@@ -0,0 +1,61 @@
...
+  // If update is available, show the latest version beside the current version in parentheses, else show the current version
+  const line = updateAvailable
+    ? `${header} ${chalk.dim("(update available")} ${chalk.green(latestVersion)}${chalk.dim(")")}`
+    : header;
+
+  console.log(line);
+
+  const columns = process.stdout.columns || 80;
+  const width = Math.max(Math.min(columns, 100), 20);
+  console.log(chalk.yellow("─".repeat(width)));
```
zwpaper: how about only showing the update informations only when updateAvailable? we can return early if updateAvailable is false.

packages/cli/src/upgrade/cmd.ts
```diff
@@ -15,14 +13,10 @@ export function registerUpgradeCommand(program: Command) {
       console.log("Checking for updates...");
 
       try {
-        const latestRelease = await fetchLatestCliRelease();
-        const latestVersion = extractVersionFromTag(latestRelease.tag_name);
-        const currentVersion = packageJson.version;
+        const { updateAvailable, latestVersion, latestRelease } =
+          await returnVersionInfo({ timeoutMs: null });
 
-        console.log(`Current version: ${currentVersion}`);
```
zwpaper, line 22: Looks like `Current version` is deleted, I would prefer also printing the current version with the latestVersion

zwpaper (changes requested):
Hi @DESU-CLUB , there are some UX comments and nits, please have a look and ping me if you have any questions

packages/cli/src/upgrade/version-check.ts
DESU-CLUB: Small clarification here, I think the `packageJson.version` should be for currentVersion, not latestRelease

currentVersion reflects the current version of Pochi while latestRelease is the latest github release fetched during `checkForUpdate`

DESU-CLUB: Also I think there are syntax errors when I try defining latestVersion and currentVersion within the return statement, as updateAvailable requires those functions.

One way to resolve it would be to call the extractVersionFromTag twice (once for latest version and updateAvailable), but it duplicates the function call

As of now here is my current fix:
```typescript
  const latestVersion = extractVersionFromTag(latestRelease.tag_name);

  return {
    updateAvailable: isNewerVersion(latestVersion, packageJson.version),
    currentVersion: packageJson.version,
    latestVersion,
    latestRelease,
  };
```

DESU-CLUB:
@zwpaper Thank you for the feedback! Have resolved the nits, and left comments for clarifications on one of the comments!

packages/cli/src/upgrade/version-check.ts
zwpaper: Oops, you are right, the packageJson.version was a typing mistake from me

packages/cli/src/upgrade/cmd.ts
```diff
@@ -15,14 +13,15 @@ export function registerUpgradeCommand(program: Command) {
...
+          updateAvailable,
+          latestVersion,
+          latestRelease,
+          currentVersion,
+        } = await returnVersionInfo({ timeoutMs: null });
         console.log(`Current version: ${currentVersion}`);
-        console.log(`Latest version: ${latestVersion}`);
 
-        if (isNewerVersion(latestVersion, currentVersion)) {
+        if (updateAvailable) {
```
zwpaper, line 24: when update is not available, we should tell user that they were already on the latest version

zwpaper (changes requested):
Hi @DESU-CLUB, one last nit, otherwise LGTM

packages/cli/src/upgrade/cmd.ts
DESU-CLUB: Hi @zwpaper , clarifying on this

I believe this is already implemented, here is the full snippet:
```typescript
if (updateAvailable) {
          console.log(
            chalk.green(`A new version (${latestVersion}) is available!`),
          );
          await downloadAndInstall(latestRelease);
        } else {
          console.log(chalk.green("You are already on the latest version."));
        }
```

zwpaper: Oops, my fault, the code has been collapsed and I overlooked it.

PR #328, opened 2025-09-15
fix(cli): simplify binary installation process

wsxiaoys:
## Summary
- Removes the use of versioned binary files in favor of a single, consistently named binary. This change simplifies the installation and upgrade logic by eliminating the need for symlinks or version-specific filenames.

## Test plan
The pre-push hook ran the test suite, and all tests passed.


wsxiaoys:
cc @DESU-CLUB

PR #377, opened 2025-09-18
Add fun placeholder messages to empty chat screen

wsxiaoys:
## Summary
- Implement random fun placeholder messages for the empty chat screen
- Remove disclaimer text from localization files
- Add useState and useEffect to manage dynamic placeholder content
- Create array of fun placeholder messages with titles and descriptions

## Test plan
- [x] Verify that fun placeholder messages are displayed randomly on the empty chat screen
- [x] Check that the disclaimer text has been removed from all localization files
- [x] Ensure that the fallback to default text works when fun placeholders are not available
- [x] Test that the component properly handles the useEffect hook for setting placeholder content

These changes enhance the user experience by providing more engaging and fun placeholder text when the chat is empty, while removing the redundant disclaimer text.


wsxiaoys:
image

Issue #409, opened 2025-09-22
Sanitize sensitive data from shared messages in livekit-cf


wsxiaoys:
**Describe the bug**
The share functionality in `packages/livekit-cf` exposes sensitive information from tool calls in shared sessions. The endpoint responsible for generating the share data includes all messages without sanitization. Specifically, the response includes the full content of tool inputs and outputs for tools like `readFile`, `applyDiff`, and `multiApplyDiff`, which can contain sensitive code or file content.

**Link to code**
The issue is located in `packages/livekit-cf/src/do/livestore-client/app.ts`, where the JSON response for the share event is constructed. The `messages` field, populated by `inlineSubTasks(messages, subTasks)`, contains the unsanitized data.

[https://github.com/TabbyML/pochi/blob/286f93839/packages/livekit-cf/src/do/livestore-client/app.ts#L41-L59](https://github.com/TabbyML/pochi/blob/286f93839/packages/livekit-cf/src/do/livestore-client/app.ts#L41-L59)

```typescript
    return c.json({
      type: \"share\",
      messages: inlineSubTasks(messages, subTasks),
      todos: task.todos as DeepWritable<typeof task.todos>,
      isLoading: task.status === \"pending-model\",
      error: task.error,
      // ...
    } satisfies ShareEvent);
```

**Proposed Solution**
A sanitization function should be implemented and applied to the `messages` array before it's included in the response. This function should identify messages containing sensitive tool calls (e.g., `readFile`, `applyDiff`, `multiApplyDiff`) and either remove them entirely or redact the sensitive portions of their content before they are shared.

**Additional context**
This is a security and privacy concern that should be addressed to protect user data in shared sessions.

🤖 Generated with [Pochi](https://getpochi.com)

PR #418, opened 2025-09-23
chore(task): sanitize readFile, applyDiff, multiAppliDiff content in task page

zwpaper:
…task page

- readFile
  Ghostty 2025-09-24 11 48 41

- applyDiff
  Ghostty 2025-09-24 11 52 57

- multiApplyDiff
  CleanShot 2025-09-24 at 11 53 02@2x

fix https://github.com/TabbyML/pochi/issues/409

### DIFF
--- a/packages/livekit-cf/src/do/livestore-client/app.ts
+++ b/packages/livekit-cf/src/do/livestore-client/app.ts
@@ -1,7 +1,7 @@
 import { verifyJWT } from "@/lib/jwt";
 import type { ShareEvent } from "@getpochi/common/share-utils";
 import { decodeStoreId } from "@getpochi/common/store-id-utils";
-import { catalog } from "@getpochi/livekit";
+import { type Message, catalog } from "@getpochi/livekit";
 import type { ClientTools, SubTask } from "@getpochi/tools";
 import type { Store } from "@livestore/livestore";
 import type { UIMessage } from "ai";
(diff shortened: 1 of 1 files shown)


packages/livekit-cf/src/do/livestore-client/app.ts
```diff
@@ -11,6 +11,76 @@ import type { MiddlewareHandler } from "hono";
 import { HTTPException } from "hono/http-exception";
 import type { DeepWritable, Env } from "./types";
 
+const SANITIZED_MESSAGE = "[FILE CONTENT SANITIZED]";
```
wsxiaoys:
```suggestion
const REDACTED_MESSAGE = "[FILE CONTENT REDACTED]";
```

packages/livekit-cf/src/do/livestore-client/app.ts
```diff
@@ -11,6 +11,76 @@ import type { MiddlewareHandler } from "hono";
...
+
+// Sanitize sensitive content from tool calls
+function sanitizeMessage(message: UIMessage): UIMessage {
```
wsxiaoys: use `Message` from @getpochi/livekit so you get strong type of readFIle, applyDiff, etc, and doesn't have to do typeof check

wsxiaoys (approved):

PR #420, opened 2025-09-23
feat(vscode): allow open Pochi in individual tab

Sma1lboy:
https://jam.dev/c/e87823a2-36ff-4669-9f12-48c3372debfd

image

This pull request introduces a significant refactor to the VS Code extension's webview architecture. The main change is the extraction of a new `BaseWebview` abstract class to unify and simplify the logic for both sidebar and editor tab webviews. This enables the introduction of a new "Open in Editor" feature, allowing the Pochi web UI to be opened as an editor tab in addition to the sidebar. The refactor also improves dependency injection, session management, and code organization for better maintainability and extensibility.

The most important changes are:

**Webview Architecture Refactor:**

* Added a new `BaseWebview` abstract class that encapsulates shared logic for webview HTML generation, thread management, authentication event handling, and session management, to be used by both sidebar and editor tab webviews (`base-webview.ts`).
* Refactored the sidebar webview provider: renamed `RagdollWebviewProvider` to `RagdollWebviewSidebar`, making it extend `BaseWebview` and updating its implementation to use the new shared logic (`ragdoll-webview-provider.ts`).
* Updated dependency injection and references throughout the extension to use the new `RagdollWebviewSidebar` class instead of the old provider. [[1]](diffhunk://#diff-b07d9e1ce11d8e806c911933f186c2955f9d8e94aaea85cf349cc98a8a088255L15-R16) [[2]](diffhunk://#diff-b07d9e1ce11d8e806c911933f186c2955f9d8e94aaea85cf349cc98a8a088255R59-R65) [[3]](diffhunk://#diff-a776cb339ae90b33ce67c4bf373b6eb2898ab711805e99943e20951ffd9d0b87R5-R7) [[4]](diffhunk://#diff-a776cb339ae90b33ce67c4bf373b6eb2898ab711805e99943e20951ffd9d0b87L43-R52)

**New Feature: Open Pochi Web UI in Editor Tab:**

* Introduced the `PochiWebviewPanel` class, which extends `BaseWebview` and manages webviews as editor tabs. This allows users to open multiple independent Pochi panels as editor tabs, each with its own session and state (`pochi-webview-panel.ts`).
[...]

PR #435, opened 2025-09-25
Revert "feat(vscode): allow open Pochi in individual tab"

wsxiaoys:
Reverts TabbyML/pochi#420


wsxiaoys:
Reverting due to concerns around non same origin webview between sidebar and individual panel

PR #438, opened 2025-09-26
refactor(vscode): get cwd from VSCodeHostImpl

wsxiaoys:


packages/vscode/src/integrations/webview/vscode-host-impl.ts
```diff
@@ -106,6 +106,7 @@ export class VSCodeHostImpl implements VSCodeHostApi, vscode.Disposable {
   private checkpointGroup = runExclusive.createGroupRef();
   private sessionState: SessionState = {};
   private disposables: vscode.Disposable[] = [];
+  private cwd = getWorkspaceFolder()?.uri.fsPath;
```
wsxiaoys:
```suggestion
  // cwd === null means no workspace is currently open.
  private cwd: string | null = getWorkspaceFolder()?.uri.fsPath;
```

packages/cli/src/tools/index.ts
```diff
@@ -46,6 +46,7 @@ export async function executeToolCall(
         messages: [],
         toolCallId: tool.toolCallId,
         abortSignal,
+        cwd: "", // no cwd in CLI context
```
wsxiaoys: https://github.com/TabbyML/pochi/blob/main/packages/cli/src/tools/list-files.ts

we should migrate context.cwd to this ced?

ZHallen122:
Verified

Date: 10/03/2025

production: 0.8.69

PR #444, opened 2025-09-26
chore: update livestore dependencies to version 0.4.0-dev.10 and apply…

wsxiaoys:
… related patches

- Updated livestore packages in package.json from version 0.4.0-dev.7 to 0.4.0-dev.10.
- Modified LiveStoreClientDO to use new durable object structure and sync backend stub.
- Refactored worker.ts to utilize new sync request handling and validation logic.
- Added patch for create-store.js to change log level from Debug to Error.


wsxiaoys:
we probably wanna wait for dev.10 as it'll include official implementation of using do-sqlite as events storage

https://github.com/livestorejs/livestore/commit/b1e7f8c9dd24513995da7f57dd9d251cd04c7b5a

PR #446, opened 2025-09-26
refactor(vscode): update pochi code completion client to use direct API

wsxiaoys:
## Summary
- Remove unused FIM API types from pochi-api.ts
- Update Pochi code completion client to use direct API calls instead of ApiClient
- Switch to using getVendor for credentials
- Update request/response types to match Mistral API format
- Change maxTokens to max_tokens to match API expectations

## Test plan
- [x] Verify code completion still works with Pochi provider
- [x] Check that API calls are properly formatted
- [x] Ensure error handling is appropriate


wsxiaoys:
hold as we need more granular level control for code completion (reporting, logging and quotaing)

PR #464, opened 2025-09-29
refactor(persist): use webhook instead of persist api to save task to…

zwpaper:
… server and save `p-` prefixed shareId

This PR deleted the persist request, should be merged after enabling Webhook in livekit-cf

### DIFF
--- a/packages/livekit-cf/src/do/livestore-client/do.ts
+++ b/packages/livekit-cf/src/do/livestore-client/do.ts
@@ -1,14 +1,10 @@
 import { DurableObject } from "cloudflare:workers";
-import { getServerBaseUrl } from "@/lib/server";
 import { WebhookDelivery } from "@/lib/webhook-delivery";
 import type { ClientDoCallback, Env, User } from "@/types";
-import type { PochiApi, PochiApiClient } from "@getpochi/common/pochi-api";
-import { decodeStoreId } from "@getpochi/common/store-id-utils";
-import { type Task, catalog } from "@getpochi/livekit";
+import { catalog } from "@getpochi/livekit";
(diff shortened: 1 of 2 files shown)


packages/livekit-cf/src/do/livestore-client/do.ts
```diff
@@ -143,52 +132,4 @@ export class LiveStoreClientDO
...
-    const { shareId } = await resp.json();
-    if (!task.shareId) {
-      store.commit(
-        catalog.events.updateShareId({
-          id: task.id,
-          shareId,
-          updatedAt: new Date(),
-        }),
-      );
-    }
```
wsxiaoys, line 179: we shall keep the share logic update id logic using new format in onTaskUpdate

packages/livekit/src/livestore/schema.ts
```diff
@@ -195,6 +195,7 @@ const materializers = State.SQLite.materializers(events, {
       id,
       status: initMessage ? "pending-model" : "pending-input",
       parentId,
+      shareId: `p-${id.replaceAll("-", "")}`,
```
wsxiaoys:
```suggestion
```

ShareId update shall still happens in livekit-cf

wsxiaoys:
/pochi please address comments and update PR

packages/livekit/src/livestore/schema.ts
```diff
@@ -195,6 +195,7 @@ const materializers = State.SQLite.materializers(events, {
       id,
       status: initMessage ? "pending-model" : "pending-input",
       parentId,
+      shareId: null,
```
wsxiaoys:
```suggestion
```

PR #508, opened 2025-10-07
feat: re-enable workspace configuration handling

wsxiaoys:


packages/vscode/src/integrations/command.ts
```diff
@@ -431,7 +433,13 @@ export class CommandManager implements vscode.Disposable {
 
       vscode.commands.registerCommand("pochi.openInPanel", async () => {
         // FIXME(zhanba): pass cwd from command argument
```
wsxiaoys, line 435: @zhanba fyi

PR #517, opened 2025-10-09
feat(workflow): support global workflows

### LINKED ISSUE #123: feat(workflows): Support global workflows

Currently, Pochi supports global rules which can be defined in `~/.pochi/README.pochi.md`. This is a great feature for setting up consistent behavior across different projects. However, workflows are only loaded from the current workspace's `.pochi/workflows` directory.

This feature request is to add support for global workflows, which would be loaded from `~/.pochi/workflows`. This would allow users to define a set of common workflows that they use across all of their projects, without having to copy them into each project's workspace.

The implementation should be similar to how global rules are handled in [`packages/common/src/tool-utils/custom-rules.ts`](https://github.com/TabbyML/pochi/blob/main/packages/common/src/tool-utils/custom-rules.ts). The workflow collection logic in [`packages/vscode/src/lib/env.ts`](https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/lib/env.ts) should be updated to look for workflows in both the workspace and the global `~/.pochi/workflows` directory.

**Additional context**

This would improve the user experience by making it easier to manage and reuse common workflows.

liangfung:
## Screenshot
image

## Docs
image

Closes #123

### DIFF
--- a/packages/vscode/src/lib/__test__/env.test.ts
+++ b/packages/vscode/src/lib/__test__/env.test.ts
@@ -596,32 +596,31 @@ describe("env.ts", () => {
       assert.strictEqual(workflows[0].id, "workspace-workflow");
     });
 
-    it("should handle duplicate workflow names from workspace and global directories", async () => {
-      // Create workspace workflow with same name
-      await createDirectory(workflowsDir);
-      const workspaceWorkflowContent = "# Workspace Version\nThis is the workspace version";
-      await createFile(vscode.Uri.joinPath(workflowsDir, "duplicate-workflow.md"), workspaceWorkflowContent);
-
-      // Create global workflow with same name
+    it("should prioritize workspace workflows over global ones with the same name", async () => {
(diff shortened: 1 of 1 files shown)


packages/vscode/src/lib/env.ts
```diff
@@ -90,42 +91,95 @@ export async function collectCustomRules(
...
+
+            const content = await readFileContent(absolutePath);
+            const frontmatter = await parseWorkflowFrontmatter(content);
+            // e.g., "workflow1.md" -> "workflow1"
+            const fileName = name.replace(/\.md$/i, "");
+
+            // For global workflows, replace home directory with ~
+            let file: string;
+            if (isGlobal) {
+              file = absolutePath.replace(systemInfo.homedir, "~");
```
zhanba, line 158: Can we have real path to open and label for display? in this way, don't need to change `openFile`

PR #523, opened 2025-10-09
fix(cli): turn LiveKit sync by default

zwpaper:
Currently the livekit sync is not turned on by default,
making the tasks NOT sync to server by default, which would:
1. cli task is not show in website
2. task from slack is not updated

This PR turn it on by default and can be off by setting the env.

## Summary
- Change the environment variable check from `POCHI_LIVEKIT_SYNC_ON` to `POCHI_LIVEKIT_SYNC_OFF` for better developer experience
- This makes the default behavior to have sync enabled, and only disable it when explicitly requested via the environment variable
- Fixes the logic for checking whether LiveKit sync is enabled/disabled

## Test plan
- [x] Verify that LiveKit sync works by default
- [x] Verify that setting `POCHI_LIVEKIT_SYNC_OFF=1` disables sync
- [x] All existing tests pass

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -397,7 +397,7 @@ async function waitForSync(
   inputStore?: Store,
   timeoutDuration: Duration.DurationInput = "1 second",
 ) {
-  if (!process.env.POCHI_LIVEKIT_SYNC_ON) {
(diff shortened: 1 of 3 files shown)


wsxiaoys:
I recommend set POCHI_LIVEKIT_SYNC_ON=true in slack environment instead.

PR #525, opened 2025-10-10
fix(livekit): properly handle the conversion of tool results of JSON type

liangfung:


liangfung:
In the CLI, an unexpected `toolResult` field with the value `undefined` is added to the mcp toolResult output, causing a validation failure when calling `addToolResult` and resulting in a program error. This field needs to be addressed

zwpaper:
verified with multiple MCP Toolcalls:

CleanShot 2025-10-23 at 15 41 51@2x

PR #537, opened 2025-10-11
fix(vscode): use fsPath in workspace job queue

zwpaper:
the workspace job queue filter is updated to use fsPath in:

https://github.com/TabbyML/pochi/pull/438/files#diff-dfe98539e5a6eda6aa68e37e7da4b2bfc245122dd180bc7ce5dbbe67df405732

so we should update the ones in uri-handler

### DIFF
--- a/packages/vscode/src/lib/workspace-job.ts
+++ b/packages/vscode/src/lib/workspace-job.ts
@@ -3,7 +3,7 @@ import { inject, injectable, singleton } from "tsyringe";
 import * as vscode from "vscode";
 
 export interface WorkspaceJob {
-  // should get from vscode.Uri.toString()
+  // should get from vscode.Uri.fsPath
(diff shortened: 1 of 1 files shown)


packages/vscode/src/integrations/uri-handler.ts
```diff
@@ -98,7 +98,7 @@ class RagdollUriHandler implements vscode.UriHandler, vscode.Disposable {
 
     // Push job to open the task after workspace opens
     await this.workspaceJobQueue.push({
-      workspaceUri: existingProject.toString(),
+      workspaceUri: existingProject.fsPath,
```
icycodes, line 101: Please also update comment in https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/lib/workspace-job.ts#L6

PR #559, opened 2025-10-16
fix(config): use rename to make sure config atomic write

zwpaper:


wsxiaoys:
Please verify the behavior works as expected for config watchers

zwpaper:
fs.watch listen on the file handler, it can not response on rename.

so I have changed to watch the dir and response on config file changes, tested locally with login/logout, i can see the callback log during my local test on every file change steps.

packages/common/src/configuration/config-file.ts
```diff
@@ -95,7 +100,14 @@ export class PochiConfigFile {
       });
       content = JSONC.applyEdits(content, edits);
 
-      await fsPromise.writeFile(this.configFilePath, content);
+      // writeFile is not atomic,
+      // the writing of multiple instance of pochi in the same time would break the config
+      // so we use rename to make sure the config remains valid during write
+      //
+      // the caveat is only the last writer wins, but it's acceptable
+      const tmp = `${this.configFilePath}.${randomUUID()}.tmp`;
```
wsxiaoys: use node mktemp api

wsxiaoys (commented):
otherwise LG

wsxiaoys:
/pochi address comments from `gh api repos/foo/bar/pulls/123/comments`

PR #569, opened 2025-10-17
feat: allow readFile tool to read images for vision-capable models

### LINKED ISSUE #539: feat(tools): allow readFile to read images for supported models
**Is your feature request related to a problem? Please describe.**
Currently, the `readFile` tool only supports reading text files. It would be beneficial to extend its functionality to allow reading image files for models that support image input (e.g., Gemini, Anthropic).

**Describe the solution you'd like**
I propose updating the `readFile` tool to detect the file type based on its extension or MIME type. If the file is an image, the tool should read it as a base64-encoded string and return it in a format that can be consumed by multimedia-capable models.

The implementation could be similar to how image outputs are handled in `packages/livekit/src/chat/mcp-utils.ts`.

Specifically, the `outputSchema` of the `readFile` tool in `packages/tools/src/read-file.ts` could be updated to support a content union type, similar to the `ContentOutput` in `mcp-utils.ts`:

```typescript
const ContentOutput = z.union([
  z.object({
    ...
  }),
]);
```

**Describe alternatives you've considered**
An alternative would be to create a new tool specifically for reading images, but extending the existing `readFile` tool seems more intuitive and efficient.

**Additional context**
This feature would enhance Pochi's ability to work with multimodal models and enable more complex interactions involving images.

Relevant files:
* [packages/tools/src/read-file.ts](https://github.com/TabbyML/pochi/blob/main/packages/tools/src/read-file.ts)
* [packages/livekit/src/chat/mcp-utils.ts](https://github.com/TabbyML/pochi/blob/main/packages/livekit/src/chat/mcp-utils.ts)

zhanba:
## Screenshot

image

image

## Summary
- Extends the readFile tool to support reading image files for vision-capable AI models
- Adds MIME type detection and validation for media files
- Returns images as base64-encoded data with proper media type information
- Updates sanitize-message to handle media content types correctly
- Adds comprehensive test coverage for type safety with union return types

## Implementation Details
- Added `isSupportedMediaFile` utility function to validate file types based on MIME types
- Modified readFile to return a union type supporting both text and media content
- Updated message sanitization logic to properly handle media types instead of just image types
- Added type guards in tests to ensure type safety when checking content

## Related Issue
Fixes #539

## Test Plan
- [x] Unit tests pass for readFile with text files
- [x] Unit tests pass for readFile with image files
- [x] Type checking works correctly with union return types
- [x] Pre-push hook tests pass successfully

### DIFF
--- a/packages/tools/src/read-file.ts
+++ b/packages/tools/src/read-file.ts
@@ -1,7 +1,7 @@
 import { z } from "zod";
 import { defineClientTool } from "./types";
 
-const TextOutputSchema = z.object({
+const TextOutput = z.object({
   type: z.literal("text").optional(),
   content: z.string(),
   isTruncated: z
@@ -11,11 +11,10 @@ const TextOutputSchema = z.object({
     ),
 });
 
-const MediaOutputSchema = z.object({
+export const MediaOutput = z.object({
   type: z.literal("media"),
   data: z.string().describe("The base64-encoded media data"),
   mimeType: z.string().describe("The MIME type of the media"),
-  isTruncated: z.boolean().describe("Whether the media is truncated"),
 });
 
 export const createReadFileTool = (contentType?: string[]) =>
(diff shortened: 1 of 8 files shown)


packages/tools/src/read-file.ts
```diff
@@ -1,36 +1,67 @@
...
-};
+const TextOutputSchema = z.object({
+  type: z.literal("text").optional(),
+  content: z.string(),
+  isTruncated: z
+    .boolean()
+    .describe(
+      "Whether the textual content is truncated due to exceeding the maximum length",
+    ),
+});
```
wsxiaoys: please always make schema backward compatible, otherwise messages in database won't able to render.

tips: use z.union(...)

packages/vendor-pochi/src/vendor.ts
```diff
@@ -52,6 +52,9 @@ export class Pochi extends VendorBase {
             contextWindow: x.contextWindow,
             useToolCallMiddleware: x.id.includes("google/"),
             label: x.costType === "basic" ? "swift" : "super",
+            supportedMimeTypes: x.id.includes("google/")
```
wsxiaoys: also video and pdf?

Issue #551, opened 2025-10-14
feat(cli): add graceful shutdown for stop signals (SIGINT)


icycodes:
The `pochi` CLI should gracefully shut down when it receives stop signals like `SIGINT` (e.g., when the user presses `Ctrl+C`). This ensures that resources are properly cleaned up before the process exits.

Currently, interrupting the CLI with `Ctrl+C` might lead to an abrupt termination without proper cleanup.

We should implement a signal handler for `SIGINT` in the main CLI entry point to trigger a graceful shutdown sequence.

**Relevant Code:**

*   [`packages/cli/src/cli.ts`](https://github.com/TabbyML/pochi/blob/main/packages/cli/src/cli.ts): The main entry point for the CLI.
*   [`packages/cli/src/lib/shutdown.ts`](https://github.com/TabbyML/pochi/blob/main/packages/cli/src/lib/shutdown.ts): Contains the `shutdownStoreAndExit` function, which should be called during graceful shutdown.

**Implementation Plan:**

1.  In `packages/cli/src/cli.ts`, add a `process.on('SIGINT', ...)` handler.
2.  The handler should call `shutdownStoreAndExit` and perform any other necessary cleanup before exiting.

A similar implementation for `SIGTERM` can be found in `packages/common/src/message-utils/__tests__/message-utils.test.ts`.

---
🤖 Generated with [Pochi](https://getpochi.com)

PR #573, opened 2025-10-17
feat(cli): Add graceful shutdown handling for SIGINT/SIGTERM signals

ZHallen122:
#551 
## Summary
This PR introduces graceful shutdown handling for SIGINT and SIGTERM signals in the Pochi CLI. When the process receives these termination signals, it will now properly clean up resources before exiting.

## Test plan
 - Verify CLI still functions normally
 - Test SIGINT (Ctrl+C) interruption works properly
 - Test SIGTERM signal handling
 - Confirm proper exit codes are returned
 - Verify no resource leaks during shutdown

🤖 Generated with Pochi


packages/cli/src/cli.ts
```diff
@@ -65,6 +65,34 @@ import { checkForUpdates, registerUpgradeCommand } from "./upgrade";
 const logger = getLogger("Pochi");
 logger.debug(`pochi v${packageJson.version}`);
 
+// Set up graceful shutdown for SIGINT and SIGTERM
```
wsxiaoys: how about just extract a function like createAbortControllerWithGracefulShutdown， and setup callback internally?

packages/cli/src/cli.ts
```diff
@@ -65,6 +65,44 @@ import { checkForUpdates, registerUpgradeCommand } from "./upgrade";
...
+    return () => {
+      if (isShuttingDown) return;
+      isShuttingDown = true;
+
+      if (!abortController.signal.aborted) {
+        abortController.abort(new Error(`Process interrupted by ${signal}`));
+      }
+
+      // Force exit immediately with appropriate code
+      process.exit(exitCode);
```
wsxiaoys: you should not call process.exit here, otherwise the main loop doesn't got a chance to exit gracefully

packages/cli/src/task-runner.ts
```diff
@@ -102,6 +107,7 @@ export class TaskRunner {
   private llm: LLMRequestData;
   private toolCallOptions: ToolCallOptions;
   private stepCount: StepCount;
+  private abortSignal?: AbortSignal;
```
wsxiaoys: remove

packages/livekit/src/chat/live-chat-kit.ts
wsxiaoys: revert changes in this file.

packages/cli/src/task-runner.ts
```diff
@@ -194,6 +202,11 @@ export class TaskRunner {
       logger.trace("Start step loop.");
       this.stepCount.reset();
       while (true) {
+        // Check if we should exit due to abort signal
+        if (this.abortSignal?.aborted) {
+          break;
+        }
+
```
wsxiaoys: is this still necessary? we'll exit after step anyway?

ZHallen122: You're right. I tested this and confirmed the exit after step makes this check unnecessary.

packages/cli/src/lib/shutdown.ts
```diff
@@ -1,5 +1,40 @@
...
+  const abortController = new AbortController();
+  let isShuttingDown = false;
+
+  const handleShutdown = (signal: string, _exitCode: number) => {
+    return () => {
+      if (isShuttingDown) return;
+      isShuttingDown = true;
+
+      if (!abortController.signal.aborted) {
+        abortController.abort(new Error(`Process interrupted by ${signal}`));
```
icycodes: I think the `isShuttingDown` variable is not required here, it equals `abortController.signal.aborted`.

PR #585, opened 2025-10-20
feat(livekit-cf): send task result in webhook

zwpaper:
This is safe to be merged before the PR in Ragdoll https://github.com/TabbyML/ragdoll/pull/1608

### DIFF
--- a/packages/livekit-cf/src/do/livestore-client/do.ts
+++ b/packages/livekit-cf/src/do/livestore-client/do.ts
@@ -145,7 +145,7 @@ export class LiveStoreClientDO
     if (webhook) {
       await Promise.all(
         updatedTasks.map((task) => {
-          let completionInfo: string | undefined = undefined;
+          let completion: string | undefined = undefined;
           let followup = undefined;
           if (task.status === "completed") {
             const messages = store.query(
@@ -165,7 +165,7 @@ export class LiveStoreClientDO
                     part.type === "tool-attemptCompletion" &&
                     part.state === "input-available"
                   ) {
-                    completionInfo =
+                    completion =
                       (part.input as { result?: string } | undefined)?.result ||
                       undefined;
                     break;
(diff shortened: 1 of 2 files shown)


packages/livekit-cf/src/do/livestore-client/do.ts
```diff
@@ -9,6 +9,7 @@ import type { CfTypes } from "@livestore/sync-cf/common";
 import moment from "moment";
 import { funnel } from "remeda";
 import * as runExclusive from "run-exclusive";
+import { makeMessagesQuery } from "../../../../livekit/src/livestore/queries";
```
wsxiaoys: use `catalog.queries.makeMessagesQuery`

packages/livekit-cf/src/lib/webhook-delivery.ts
```diff
@@ -6,7 +6,16 @@ export class WebhookDelivery {
...
 
-  async onTaskUpdated(task: Task) {
+  async onTaskUpdated(
+    task: Task,
+    result?: {
+      completionInfo?: string;
+      followup?: {
+        question: string;
+        followup?: string[];
+      };
```
wsxiaoys: update naming as comments in ragdoll repo

zwpaper: WIP

packages/livekit-cf/src/do/livestore-client/do.ts
```diff
@@ -144,9 +144,58 @@ export class LiveStoreClientDO
...
+                      (part.input as { result?: string } | undefined)?.result ||
+                      undefined;
+                    break;
+                  }
+                  if (
+                    part.type === "tool-askFollowupQuestion" &&
+                    part.state === "input-available"
+                  ) {
+                    followup = part.input as
+                      | { question: string; followUp?: string[] }
```
zwpaper, line 178: `followUp` comes from toolcall result, should we also change it to `choices` or `followup`? @wsxiaoys

PR #600, opened 2025-10-24
feat(vscode-webui): Add visual diff highlighting for file edits

zhanba:
## Screenshot

image

## Summary
- Added syntax highlighting for diff content using shiki transformers
- Created new ModelEdits component to show AI-generated changes
- Updated applyDiff, multiApplyDiff, and writeToFile tools to display edits
- Added CSS styles for diff highlighting in code blocks
- Modified tool call lifecycle to handle non-interactive mode for pane webview

## Test plan
- [x] Verify that diff highlighting works correctly in the UI
- [x] Test that both user edits and model edits are displayed properly
- [x] Confirm that the non-interactive mode works for pane webview
- [x] Check that all existing tests pass

### DIFF
--- a/packages/tools/src/constants.ts
+++ b/packages/tools/src/constants.ts
@@ -33,7 +33,7 @@ export const EditFileOutputSchema = z.object({
 
   _meta: z
     .object({
-      edits: z
+      edit: z
         .string()
         .describe("The diff representing the edits made to the file.")
         .optional(),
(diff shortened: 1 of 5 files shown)


packages/tools/src/constants.ts
```diff
@@ -12,6 +12,11 @@ export const EditFileOutputSchema = z.object({
     .boolean()
     .describe("Indicates whether the file was successfully written."),
 
+  edits: z
```
wsxiaoys: this should goes to _meta field, as it's for display only (and not intended to be viewed by model)

wsxiaoys:
Otherwise LGTM

PR #615, opened 2025-10-28
feat: add worktree setup script configuration

zhanba:
## Summary
This PR adds support for configuring a setup script that runs automatically after creating a new worktree in VSCode. This is useful for projects that need to run initialization commands in new worktrees.

## Changes
- **Configuration**: Added `worktree.setupScript` configuration option
  - Can be set in workspace config (`.pochi/config.jsonc`)
  - Schema updated to include the new option
  - ConfigManager updated to allow worktree config at workspace level
  
- **VSCode Integration**: Implemented setup script execution flow
  - After creating a new worktree via VSCode command, the configured script is executed in a terminal
  - Terminal opens automatically with the setup script running in the new worktree directory
  
- **Example Config**: Added `.pochi/config.jsonc` with example setup script (`sh .pochi/init.sh`)

## Usage
Add to your workspace `.pochi/config.jsonc`:
```json
{
  "worktree": {
    "setupScript": "npm install && npm run build"
  }
}
```

When creating a new worktree through VSCode, the script will automatically run in the new worktree directory.

## Test Plan
- [ ] Verify worktree creation without setupScript works as before
- [ ] Configure a setupScript and verify it executes after worktree creation
- [ ] Verify script runs in correct directory (the new worktree path)
- [ ] Verify terminal opens showing the script execution

### DIFF
--- a/packages/vscode/src/integrations/command.ts
+++ b/packages/vscode/src/integrations/command.ts
@@ -18,7 +18,7 @@ import { PostHog } from "@/lib/posthog";
 import { workspaceScoped } from "@/lib/workspace-scoped";
 // biome-ignore lint/style/useImportType: needed for dependency injection
 import { NESDecorationManager } from "@/nes/decoration-manager";
-import type { WebsiteTaskCreateEvent } from "@getpochi/common";
+import { type WebsiteTaskCreateEvent, toErrorMessage } from "@getpochi/common";
 import {
   type CustomModelSetting,
   type McpServerConfig,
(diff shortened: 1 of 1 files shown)


packages/common/src/configuration/types.ts
```diff
@@ -40,5 +40,13 @@ export function makePochiConfig(strict = false) {
     vendors: looseRecord(VendorConfig.nullable(), strict).optional(),
     providers: looseRecord(CustomModelSetting, strict).optional(),
     mcp: looseRecord(McpServerConfig, strict).optional(),
+    worktree: z
```
wsxiaoys: let's not adding this new configuration, instead, just utilize `~/.pochi/init.sh` in workspace. it's a practice we used for slack integration anyway

PR #633, opened 2025-10-31
feat(vscode-webui): sanitize livestore storeId with cwd

wsxiaoys:
## Summary
- Sanitize the `storeId` used by `LiveStoreTaskProvider` in `vscode-webui` by incorporating the current working directory (cwd).
- Ensures that the `storeId` is unique per workspace, preventing conflicts when multiple VSCode instances are open with different workspaces.
- Introduces a new `sanitizeStoreId` function to clean the cwd string, allowing only alphanumeric characters, underscores, and hyphens.

## Test plan
- Verify that tasks are correctly stored and retrieved when opening multiple VSCode instances with different workspaces.
- Check that the `storeId` generated is unique for each workspace and contains only allowed characters.

### DIFF
--- a/packages/vscode-webui/src/routes/index.tsx
+++ b/packages/vscode-webui/src/routes/index.tsx
@@ -172,7 +172,7 @@ function App() {
     return;
   }
 
-  if (!currentWorkspace) {
+  if (!currentWorkspace?.cwd) {
     return (
       <div className="flex h-screen w-full flex-col items-center justify-center">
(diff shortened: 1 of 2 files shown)


packages/vscode-webui/src/routes/index.tsx
```diff
@@ -172,7 +172,7 @@ function App() {
     return;
   }
 
-  if (!currentWorkspace) {
+  if (!currentWorkspace?.cwd) {
```
wsxiaoys, line 175: @zhanba fyi

PR #700, opened 2025-11-14
chore: update GPT 5 to GPT 5.1

zwpaper:


packages/vendor-codex/src/model.ts
```diff
@@ -92,7 +92,7 @@ function createCodexResponsesModel(
     fetch: customFetch as typeof fetch,
   });
 
-  return openai.responses(modelId || "gpt-5");
+  return openai.responses(modelId || "gpt-5.1-codex");
```
zwpaper, line 95: please note the default model here is also changed to gpt-5.1-codex

PR #710, opened 2025-11-17
feat(docs): update tab completion documentation with new features

icycodes:
## Summary
- Add documentation for diff preview and ghost text suggestion modes
- Update keyboard shortcuts and remove outdated ones
- Add section on feature conflicts with other extensions
- Remove outdated model configuration section
- Add new images to illustrate tab completion features

### DIFF
--- a/packages/docs/content/docs/tab-completion.mdx
+++ b/packages/docs/content/docs/tab-completion.mdx
@@ -37,11 +37,7 @@ Works across popular programming languages including:
 
 ## Using Tab Completion
 
-### Basic Usage
-1. Start typing your code
-2. Pochi will show edit suggestions
-3. Press `Tab` to accept the suggestion
-4. Press `Escape` to dismiss the suggestion
+Just start typing. Pochi will surface suggestions automatically. The more you use it, the more it understands your codebase, thus suggesting better edits.
 
 ### Keyboard Shortcuts
 | Key | Action |


packages/docs/content/docs/tab-completion.mdx
```diff
@@ -31,7 +39,7 @@ Works across popular programming languages including:
 
 ### Basic Usage
 1. Start typing your code
```
mufassirkazi: let's avoid bullet because we used them earlier. We can do something like:

Just start typing. Pochi will surface suggestions automatically. The more you use it, the more it understands your codebase, thus suggesting better edits.

packages/docs/content/docs/tab-completion.mdx
```diff
@@ -12,8 +12,16 @@ Pochi uses a state-of-the-art model that adapts to your coding patterns in real
 drawing on recent edits, diagnostics, and surrounding context to generate helpful,
```
mufassirkazi, line 12: Maybe mention it’s hosted and optimized at the beginning itself?

mufassirkazi: Also would be great to:

- Add a quick clariying sentence that tab completion is Inline completion + NES (so that users who search for NES will know its available) 
- It is enabled by default and do not need any special configurations

gyxlucy: +1 to clarify that there are two kinds of code completions we are offering

gyxlucy (changes requested):
Otherwise LGTM

packages/docs/content/docs/tab-completion.mdx
icycodes: Actually, we only use the NES model in the current version. The combination of using FIM model and NES model is in the roadmap and may be available in the next version.

However, the example images are intended to show the two display types (diff preview and ghost text), not the model type. 

NES (next edit suggestion) and FIM (fill in middle) refer to the model type, or rather, suggestion providers. This is not exactly equal to the display type.  
A suggestion from the NES model could also be displayed as ghost text (aka inline completion), as long as it only inserts text after the cursor. Otherwise, it will normally be displayed as a diff preview. On the other hand, a suggestion from the FIM model can be displayed as a diff preview if needed.

gyxlucy: Aligned offline with @icycodes that behind the scene, it's one model that's predicting results, but code completion or next-edit is more like different rendering methods at this moment. Let's keep things phrased loosely for the time being, and could revisit again if new updates coming in.

PR #712, opened 2025-11-17
fix(vscode-webui): improve todoWrite tool input handling

liangfung:
## Summary
- Enhanced the todoWrite tool component to properly handle different input data structures
- Implemented JSON parsing for string inputs that need to be converted to Todo arrays
- Added validation for todo objects to ensure proper structure
- Used useMemo for efficient computation of todo lists

## Screenshot
image

## Test plan
- [x] Verified component renders correctly with various input formats
- [x] Tested JSON string parsing functionality
- [x] Confirmed todo object validation works as expected
- [x] Checked that useMemo provides efficient recomputation

### DIFF
--- a/packages/vscode-webui/src/components/tool-invocation/tools/todo-write.tsx
+++ b/packages/vscode-webui/src/components/tool-invocation/tools/todo-write.tsx
@@ -1,8 +1,6 @@
 import { cn } from "@/lib/utils";
-import { Todo } from "@getpochi/tools";
 import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
-import { z } from "zod";
 import { StatusIcon } from "../status-icon";
 import { ExpandableToolContainer } from "../tool-container";
 import type { ToolProps } from "../types";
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/components/tool-invocation/tools/todo-write.tsx
```diff
@@ -10,7 +11,25 @@ export const todoWriteTool: React.FC<ToolProps<"todoWrite">> = ({
...
+      } catch (e) {
+        value = [];
+      }
+    }
+
+    if (Array.isArray(value)) {
+      return value.filter((x) => {
+        return !!x && "id" in x && "content" in x;
+      }) as Todo[];
+    }
```
wsxiaoys: Todo is defined as a zod object, you shall use Todo.safeParse instead.

https://github.com/TabbyML/pochi/blob/69c211c00ef681c1c9ee15c2e1d18a86a671d28a/packages/common/src/tool-utils/workflow-parser.ts#L38

packages/vscode-webui/src/components/tool-invocation/tools/todo-write.tsx
```diff
@@ -10,7 +11,25 @@ export const todoWriteTool: React.FC<ToolProps<"todoWrite">> = ({
...
+  const todos = useMemo(() => {
+    let value = tool.input?.todos;
+
+    if (typeof value === "string") {
+      try {
+        value = JSON.parse(value) as Todo[];
+      } catch (e) {
+        value = [];
+      }
+    }
```
wsxiaoys: why we need to do JSON.parse here? input.todos is expected to be an array object.

Use safeParse should be sufficient

wsxiaoys (changes requested):

packages/vscode-webui/src/components/tool-invocation/tools/todo-write.tsx
liangfung: When using the qwen/qwen3-coder model, having input.todos as a string type leads to a rendering error

wsxiaoys (approved):

PR #742, opened 2025-11-24
feat: persisting line changes

liangfung:


packages/livekit/src/livestore/default-schema.ts
```diff
@@ -36,6 +36,12 @@ export const tables = {
         nullable: true,
         schema: ToolCalls,
       }),
+      added: State.SQLite.integer({
```
wsxiaoys: use a json field

PR #778, opened 2025-11-28
feat(vscode-webui): improve worktree list handling when no worktrees exist

zhanba:
## Summary
- Add fallback to show current workspace as default worktree when no git worktrees exist
- Add loading state handling in WorktreeList component
- Rename repositories map to repositoryState in GitStateMonitor for clarity
- Ensure worktree order matches VSCode's repository order
- Filter out non-text files from diff display

## Test plan
- [ ] Verify that worktree list displays properly when no git worktrees exist
- [ ] Check that the current workspace is shown as the default worktree
- [ ] Test that loading states are properly handled
- [ ] Confirm that worktree ordering matches VSCode's repository order
- [ ] Verify that non-text files are properly filtered from diff display

### DIFF
--- a/packages/vscode-webui/src/components/worktree-list.tsx
+++ b/packages/vscode-webui/src/components/worktree-list.tsx
@@ -14,10 +14,12 @@ import {
   TooltipContent,
   TooltipTrigger,
 } from "@/components/ui/tooltip";
+import { useCurrentWorkspace } from "@/lib/hooks/use-current-workspace";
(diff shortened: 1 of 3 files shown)


zwpaper:
1. For binary changes, we should show the file without the content changes instead of hiding the binary directly without notice. let's keep this as is in this release and fix it later. WDYT @wsxiaoys 

Otherwise LGTM

zwpaper (approved):

PR #791, opened 2025-12-01
feat(tools): remove multiApplyDiff tool

wsxiaoys:
## Summary
- Removed the `multiApplyDiff` tool and all its references.
- Integrated the functionality of `multiApplyDiff` into `applyDiff`.

### DIFF
--- a/packages/livekit-cf/src/lib/sanitize-message.ts
+++ b/packages/livekit-cf/src/lib/sanitize-message.ts
@@ -46,6 +46,24 @@ export function sanitizeMessage(message: Message): Message {
         };
       }
 
+      if (
+        part.type === "tool-multiApplyDiff" &&
+        part.input?.edits &&
+        part.input.edits.length > 0
+      ) {
+        return {
+          ...part,
+          input: {
+            ...part.input,
+            edits: part.input.edits.map((edit: unknown) => ({
+              ...(edit as Record<string, unknown>),
+              searchContent: RedactedMessage,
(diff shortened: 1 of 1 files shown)


packages/livekit-cf/src/lib/sanitize-message.ts
```diff
@@ -46,24 +46,6 @@ export function sanitizeMessage(message: Message): Message {
         };
       }
 
-      if (
-        part.type === "tool-multiApplyDiff" &&
-        part.input?.edits &&
-        part.input.edits.length > 0
```
pingfengafei, line 52: How about
```
part.input?.edits?.length
```

packages/tools/src/index.ts
```diff
@@ -13,7 +13,7 @@ import { attemptCompletion } from "./attempt-completion";
 import { executeCommand } from "./execute-command";
 import { globFiles } from "./glob-files";
 import { listFiles } from "./list-files";
-import { multiApplyDiff } from "./multi-apply-diff";
+import type { multiApplyDiff } from "./multi-apply-diff";
```
pingfengafei, line 16: Looks like type is more likely starts with upper camel case.

packages/livekit-cf/src/lib/sanitize-message.ts
wsxiaoys, line 1: Need revert change to this file to ensure runtime is backward compatible

PR #865, opened 2025-12-15
chore: enforce useHookAtTopLevel rule in linter configuration

wsxiaoys:


packages/vscode-webui/src/lib/store-blob.ts
```diff
@@ -9,6 +9,7 @@ export function setBlobUrl(key: string, data: Blob) {
 }
 
 export function useStoreBlobUrl(inputUrl: string): string | null {
+  const { store } = useStore();
```
wsxiaoys, line 12: Have you checked if this works with share page?

PR #876, opened 2025-12-16
refactor(vscode): optimize branch name generation

wsxiaoys:
- Optimize branch name generation prompt for brevity and enforce 32-character limit


packages/vscode/src/lib/generate-branch-name.ts
icycodes, line 70: Should we also update the maxOutputToken according the MaxChars?

wsxiaoys: no need - as tokens != chars. let's leave it as is for now (64 is small enough)

PR #885, opened 2025-12-16
docs: update parallel agents documentation to include new worktree creation method

icycodes:
docs: update parallel agents documentation to include new worktree creation method


mufassirkazi (commented):
I would prefer if we can add an image after the second paragraph, but otherwise lg!

image

PR #935, opened 2025-12-24
fix: add filter completion tools middleware

icycodes:
## Summary
- Implemented `FilterCompletionToolsMiddleware` to restrict the usage of completion tools (`askFollowupQuestion` and `attemptCompletion`) when other tools are being called.
- Updated tool descriptions for `askFollowupQuestion` and `attemptCompletion` to explicitly state they cannot be combined with other tools in a single step.
- Registered the new middleware in `FlexibleChatTransport`.

## Test plan
- Verified that the middleware correctly filters out completion tools when multiple tools are present in a single request.
- Ensured the tool descriptions are updated in the generated schema.

### DIFF
--- a/packages/tools/src/ask-followup-question.ts
+++ b/packages/tools/src/ask-followup-question.ts
@@ -10,6 +10,8 @@ Use this tool in the following scenarios:
 1. The user's request is ambiguous or unclear and requires clarification.
 2. You need more details to proceed effectively.
 3. You have made several unsuccessful attempts to solve the issue and need user guidance to move forward.
+
+IMPORTANT: This tool CANNOT be used in combination with other tools (except todoWrite) in a single step. If you need to use other tools, you must do so in a separate step before calling this tool.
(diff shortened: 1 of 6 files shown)


packages/tools/src/ask-followup-question.ts
```diff
@@ -10,6 +10,8 @@ Use this tool in the following scenarios:
 1. The user's request is ambiguous or unclear and requires clarification.
 2. You need more details to proceed effectively.
 3. You have made several unsuccessful attempts to solve the issue and need user guidance to move forward.
+
+IMPORTANT: This tool CANNOT be used in combination with other tools (except todoWrite) in a single step. If you need to use other tools, you must do so in a separate step before calling this tool.
```
wsxiaoys, line 14: extract constant for reuse with askFollowUpQuestion

icycodes:
Close this PR due to #936

PR #942, opened 2025-12-24
refactor: update DiffChangesContentProvider to track change type

wsxiaoys:
- Rename static methods for clarity: `decode` -> `encode` for URI creation, `parse` -> `decode` for parsing.\n- Add `type` field to `DiffChangesData` to distinguish between 'base' and 'updated' content.\n- Update usages in command, worktree, and webview host integrations.\n- Restrict commenting in ReviewController to only 'updated' diff content.\n\n🤖 Generated with [Pochi](https://getpochi.com)

### DIFF
--- a/packages/vscode/src/integrations/editor/diff-changes-content-provider.ts
+++ b/packages/vscode/src/integrations/editor/diff-changes-content-provider.ts
@@ -5,7 +5,7 @@ interface DiffChangesData {
   filepath: string;
   content: string;
   cwd: string;
-  type: "base" | "updated";
+  type: "original" | "modified";
 }
 
 @injectable()
--- a/packages/vscode/src/integrations/review-controller.ts
+++ b/packages/vscode/src/integrations/review-controller.ts
@@ -57,7 +57,7 @@ export class ReviewController implements vscode.Disposable {
         }
 
         const changesData = DiffChangesContentProvider.decode(document.uri);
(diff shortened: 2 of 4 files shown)


packages/vscode/src/integrations/review-controller.ts
```diff
@@ -51,6 +52,15 @@ export class ReviewController implements vscode.Disposable {
...
 
+        if (document.uri.scheme !== DiffChangesContentProvider.scheme) {
+          return [];
+        }
+
+        const changesData = DiffChangesContentProvider.decode(document.uri);
+        if (changesData.type !== "updated") {
+          return [];
+        }
+
```
wsxiaoys, line 63: @zwpaper only display comment range in changes

packages/vscode/src/integrations/editor/diff-changes-content-provider.ts
```diff
@@ -5,6 +5,7 @@ interface DiffChangesData {
   filepath: string;
   content: string;
   cwd: string;
+  type: "base" | "updated";
```
zhanba:
```suggestion
  type: "original" | "modified";
```
this naming may more align with vscode

wsxiaoys: done

Issue #955, opened 2025-12-25
Review changes tab does not reflect later changes by Pochi


zwpaper:
**Describe the bug**
When the "Review changes" tab (opened via click the files in diff view) is active, it displays a snapshot of the changes at the time it was opened. If Pochi makes subsequent changes to the files, the review view does not update to reflect these new changes. Users have to close and reopen the review view to see the latest changes.

**Relevant Code**
The `showWorktreeDiff` function in `packages/vscode/src/integrations/git/worktree.ts` calculates the diff and opens the view. The content of the modified files is read at the time of execution and encoded into the URI for `DiffChangesContentProvider`.

https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/git/worktree.ts#L481-L553

Specifically, the content is baked into the URI here:
https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/git/worktree.ts#L543-L550

And the provider decodes it:
https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/editor/diff-changes-content-provider.ts#L22-L25

**Suggested Fix**
We should investigate ways to make the "Review changes" view reactive to file changes, or provide a mechanism to refresh the view when Pochi completes a task or modifies files.

**Information about your version**
Pochi VSCode Extension

---
🤖 Generated with [Pochi](https://getpochi.com)

PR #956, opened 2025-12-25
refactor(vscode): make diff editor modified content editable

zhanba:
## Summary
- Created a new unified `showDiffChanges` function and `FileChange` type to consolidate diff viewer functionality
- Replaced the `GitDiff` type with `FileChange` throughout the checkpoint system
- Added `readModifiedFromFile` parameter to make the modified side of diff editors editable when appropriate (when modified content matches disk)
- Removed duplicate `showDiff` implementation from `vscode-host-impl.ts`
- Deleted the now-unused `types.ts` file in checkpoint module

## Benefits
- **Improved UX**: Users can now edit the modified side of diff editors when viewing current file changes, allowing them to make quick adjustments without switching editors
- **Better code organization**: Consolidated two separate implementations of diff viewing logic into a single, reusable function
- **Type clarity**: Removed the unused `status` field from the type, making it clearer what data is actually used

## Test plan
- [x] All existing tests pass (157 passing)
- [ ] Manually verify that checkpoint diff views work correctly
- [ ] Manually verify that worktree diff views work correctly  
- [ ] Verify that the modified editor is editable for worktree diffs
- [ ] Verify that the modified editor is editable for checkpoint diffs when no modified commit is specified
- [ ] Verify that the modified editor is read-only for checkpoint diffs when a modified commit is specified

### DIFF
--- a/packages/vscode/src/integrations/checkpoint/checkpoint-service.ts
+++ b/packages/vscode/src/integrations/checkpoint/checkpoint-service.ts
@@ -13,8 +13,8 @@ import type {
 } from "@getpochi/common/vscode-webui-bridge";
 import { Lifecycle, inject, injectable, scoped } from "tsyringe";
 import type * as vscode from "vscode";
+import type { FileChange } from "../editor/diff-changes-editor";
 import { ShadowGitRepo } from "./shadow-git-repo";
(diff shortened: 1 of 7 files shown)


wsxiaoys:
fix #955

PR #963, opened 2025-12-26
fix: enable chat submission in non-git projects

zwpaper:
## Summary
Fixes the issue where clicking Enter in the chat sidebar has no response when working in non-git projects.

## Problem
Previously, when users tried to submit messages in the chat sidebar within non-git projects:
- `createWorktree()` would return `null` because the directory is not a git repository
- The code had a blocking `if (worktree)` condition that prevented task creation
- This resulted in no visible feedback to the user - the chatbox appeared frozen

## Changes
- Remove blocking condition that prevented task creation when worktree creation fails
- Allow task creation to proceed using current working directory for non-git projects
- Extract duplicated worktree creation logic into reusable `createWorktreeAndOpenTask` function
- Improve code maintainability by reducing ~40 lines of duplication

## Testing
- ✅ TypeScript compilation passes
- ✅ All linter checks pass
- ✅ Extension tests pass (157 passing)

## tests

Non git:

CleanShot 2025-12-26 at 14 51 09@2x

new with changing base branch
CleanShot 2025-12-26 at 14 58 17@2x

### DIFF
--- a/packages/vscode-webui/src/features/chat/components/create-task-input.tsx
+++ b/packages/vscode-webui/src/features/chat/components/create-task-input.tsx
@@ -211,31 +211,31 @@ export const CreateTaskInput: React.FC<CreateTaskInputProps> = ({
       setIsCreatingTask(true);
       setDebouncedIsCreatingTask(true);
 
+      // Upload files if present
+      let uploadedFiles: Array<{
+        contentType: string;
+        name: string;
+        url: string;
+      }> = [];
+
       if (files.length > 0) {
         const uploadedAttachments = await upload();
-        const uploadedFiles = uploadedAttachments.map((x) => ({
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/features/chat/components/create-task-input.tsx
```diff
@@ -160,52 +201,29 @@ export const CreateTaskInput: React.FC<CreateTaskInputProps> = ({
...
-        }
+        await createWorktreeAndOpenTask({
+          content,
+          shouldCreateWorktree:
+            shouldCreateWorktree === true ||
+            selectedWorktree === "new-worktree",
+          uploadedFiles,
+        });
+
+        clearFiles();
```
zhanba: should only clearFile & clearDraft if we create worktree success

zhanba (approved):

PR #981, opened 2025-12-29
docs: inline comments documentation

mufassirkazi:


packages/docs/content/docs/inline-comments.mdx
```diff
@@ -0,0 +1,67 @@
...
+    </video>
+
+
+## How to Use Inline Comments
+
+### 1. Run a Prompt
+Generate code or request changes like normal.
+
+```bash
+/pochi please use Mermaid Chart instead of Vega Lite
```
zwpaper: why adding the `/pochi` prefix?

packages/docs/content/docs/inline-comments.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+![Pochi Models](../assets/images/multiple-comments-inline-comment.png)
+
+
+### 5. Send Comments as a Batch
+Once ready:
+* Click `Submit Reviews` to Pochi
+* Or run the command: 
+  ```bash
+  /pochi please review the comments
```
zwpaper: ditto

packages/docs/content/docs/inline-comments.mdx
```diff
@@ -0,0 +1,67 @@
...
+### 4. Add Multiple Comments
+Leave comments across the file as needed. Each comment is automatically linked to its exact line + file path.
+
+![Pochi Models](../assets/images/multiple-comments-inline-comment.png)
+
+
+### 5. Send Comments as a Batch
+Once ready:
+* Click `Submit Reviews` to Pochi
+* Or run the command:
```
zwpaper: mention that this works just like submit reviews if submitting empty messages, users could add additional messages and submit in chatbox if necessary.

mufassirkazi: sg! updated step 5 to add these lines

packages/docs/content/docs/inline-comments.mdx
mufassirkazi: removed it.

gyxlucy (approved):

Issue #983, opened 2025-12-29
Enable userEdit in chatbox and prompt and fix related issues


zwpaper:
**Please describe the feature you want**
We implement userEdit functionality in v0.22, should enable it in next release and fix related issues

**Additional context**

Related issue: https://github.com/TabbyML/pochi/issues/923

---
Please reply with a 👍 if you want this feature.

PR #997, opened 2025-12-30
fix: hide user edit if Pochi made edits

zhanba:
image

## Summary
- Hides the user edit view if Pochi has made edits to the file.


zhanba:
fix #983

Issue #984, opened 2025-12-30
Show a reload button when models fail to load


liangfung:
**Please describe the feature you want**
When the model list fails to load (e.g. network issue or backend error), the user is left without models and no way to retry without reloading the entire window.
We should implement a reload/retry button in the `ModelSelect` component or the UI to allow re-fetching the model list.

**Additional context**
Relevant code:
- [`packages/vscode-webui/src/components/model-select.tsx`](https://github.com/TabbyML/pochi/blob/main/packages/vscode-webui/src/components/model-select.tsx): The component displaying the model dropdown.
- [`packages/vscode-webui/src/lib/hooks/use-model-list.ts`](https://github.com/TabbyML/pochi/blob/main/packages/vscode-webui/src/lib/hooks/use-model-list.ts): The hook responsible for fetching the model list using `useQuery`.

The `useModelList` hook uses `useQuery`, so we can potentially expose the `refetch` function and pass it down to `ModelSelect`.

🤖 Generated with [Pochi](https://getpochi.com)

PR #1000, opened 2025-12-31
feat: add reload button when model list fails to load

liangfung:
## Summary
- Show a reload button in the model selector dropdown when hosted models (Super/Swift) fail to load or are empty
- Add `refreshModelList` method to the vscode host bridge to trigger model list refresh
- Implement refresh logic in `useModelList` hook with loading state management
- Update i18n translations for reload button text across all languages
- Only display reload button for authenticated users
- Button shows spinning animation while refreshing

## Demo
https://jam.dev/c/daa7eb4d-63ed-451c-8c3e-329ac933cee8

## Test plan
- [ ] Verify reload button appears when model list fails to load
- [ ] Test that clicking the reload button triggers a refresh
- [ ] Confirm button shows spinning animation during refresh
- [ ] Verify button only appears for authenticated users
- [ ] Test that model list updates correctly after refresh

### DIFF
--- a/packages/vscode-webui/src/components/model-select.tsx
+++ b/packages/vscode-webui/src/components/model-select.tsx
@@ -195,7 +195,7 @@ export function ModelSelect({
                 ))}
 
                 {!!customModels?.flat().length && <DropdownMenuSeparator />}
-                <div className="flex items-center justify-between px-2">
+                <div className="flex items-center justify-between gap-2 px-2">
                   <a
                     href="command:pochi.openCustomModelSettings"
                     target="_blank"


liangfung:
fix #984

packages/common/src/vscode-webui-bridge/webview.ts
```diff
@@ -300,6 +300,11 @@ export interface VSCodeHostApi {
 
   readModelList(): Promise<ThreadSignalSerialization<DisplayModel[]>>;
```
wsxiaoys: no need to add an api, you can simply let readModelList return both a signal and a function (the function will trigger the reload.
readModelList() : Promise<{signal: {modelList, isLoading}, reload: () => Promise<void>}>;

zhanba (approved):

Issue #989, opened 2025-12-30
Pochi tab title should display 'workspace' for non-git workspace


liangfung:
**Describe the bug**
When Pochi is used in a workspace that is not a git repository, the Pochi tab title displays the current directory name (e.g., "my-project - 001") instead of "workspace - 001". It should display "workspace" to be consistent with the main worktree behavior in git repositories.

**Information about your version**
Current codebase.

**Additional context**
The issue seems to stem from `WorktreeManager.getWorktreeDisplayName` in `packages/vscode/src/integrations/git/worktree.ts`.

https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/git/worktree.ts#L36-L42

```typescript
  public getWorktreeDisplayName(cwd: string): string | undefined {
    const worktree = this.worktrees.value.find((wt) => wt.path === cwd);
    if (!worktree) {
      // Falls back to directory name if not a git worktree
      return getWorktreeNameFromWorktreePath(cwd);
    }
    return worktree.isMain ? "workspace" : getWorktreeNameFromWorktreePath(cwd);
  }
```

This method is used in `PochiTaskEditorProvider.createTaskUri` to generate the display name:

https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/webview/webview-panel.ts#L106-L109

```typescript
    const worktreeName = container
      .resolve(WorktreeManager)
      .getWorktreeDisplayName(params.cwd);
```

If `WorktreeManager` is not initialized (because it's not a git repo) or the cwd is not found in worktrees, it returns the folder name. We should likely check if `cwd` matches the main workspace folder and return "workspace" in that case as well.

🤖 Generated with [Pochi](https://getpochi.com)

PR #1012, opened 2026-01-02
fix(vscode): return 'workspace' when cwd matches workspace path (#989)

wsxiaoys:
## Summary
- Fixes an issue where `getWorktreeDisplayName` returned `undefined` (or fell back to path name) instead of "workspace" when the current working directory matched the workspace path but was not found in the worktrees list.
- Adds a regression test case.

## Test plan
- Verified with new test case `packages/vscode/src/integrations/git/__tests__/issue-989.test.ts`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -256,7 +256,7 @@
     },
     "packages/vscode": {
       "name": "pochi",
-      "version": "0.21.0-dev",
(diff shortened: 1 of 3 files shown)


wsxiaoys:
fix #989

liangfung (approved):

Issue #991, opened 2025-12-30
Open task tab defaults to preview mode


zhanba:
**Description**
Currently, when opening a new task tab in Pochi, it defaults to preview mode. This behavior causes the new tab to replace the current tab if the current tab is also in preview mode (which is standard VS Code behavior for preview tabs). We want to change this so that new task tabs are not opened in preview mode by default, ensuring they open as distinct tabs and do not overwrite the existing one.

**Relevant Code**
- [packages/vscode/src/integrations/layout.ts](https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/layout.ts)
  - Specifically, the `openTaskInColumn` function uses `vscode.commands.executeCommand('vscode.openWith', ...)` with `{ preview: false }`, but we should ensure this is consistently applied and effective.
  ```typescript
  async function openTaskInColumn(uri: vscode.Uri) {
    const params = PochiTaskEditorProvider.parseTaskUri(uri);
    if (!params) {
      throw new Error(`Failed to parse task URI: ${uri.toString()}`);
    }
    const viewColumn = await getViewColumnForTask(params);
    await vscode.commands.executeCommand(
      "vscode.openWith",
      uri,
      PochiTaskEditorProvider.viewType,
      { preview: false, viewColumn },
    );
  }
  ```

- [packages/vscode/src/integrations/webview/webview-panel.ts](https://github.com/TabbyML/pochi/blob/main/packages/vscode/src/integrations/webview/webview-panel.ts)
  - `PochiTaskEditorProvider.openTaskEditor` calls `openTaskInColumn`.

**Suggested Fix**
Investigate why `preview: false` might not be behaving as expected or if there are other places where the tab opening logic needs to enforce non-preview mode. Ensure that `vscode.openWith` or any other tab opening mechanism explicitly sets the preview option to false to prevent tab replacement.

🤖 Generated with [Pochi](https://getpochi.com)

PR #1020, opened 2026-01-05
fix(vscode): enable preview mode for task tab

zhanba:
## Summary
Enable preview mode when opening the task tab in VS Code. This ensures that the task tab opens as a preview tab, improving tab management for users.

## Test plan
1. Open a task in Pochi.
2. Verify that the task tab opens in preview mode (italicized title).
3. Click on another file or task; the preview tab should be replaced unless pinned or modified.


zhanba:
fix #991

Issue #986, opened 2025-12-30
URLs in CLI tasks should not be incorrectly interpreted as same-name workflows


liangfung:
**Describe the bug**
URLs in CLI tasks are incorrectly interpreted as same-name workflows.
The regex used to identify slash commands matches path segments in URLs. For example, if a user includes a URL like `https://example.com/docs` in their prompt, and there happens to be a workflow named `docs`, the `/docs` part of the URL is detected as a slash command and replaced with the workflow content.

**Relevant Code**
The issue stems from the regex used in `packages/cli/src/lib/match-slash-command.ts`:

https://github.com/TabbyML/pochi/blob/main/packages/cli/src/lib/match-slash-command.ts#L6-L14

```typescript
export function containsSlashCommandReference(prompt: string): boolean {
  return /\/\w+[\w-]*/.test(prompt);
}

export function extractSlashCommandNames(prompt: string): string[] {
  const slashCommandRegex = /(\/\w+[\w-]*)/g;
  const matches = prompt.match(slashCommandRegex);
  // ...
}
```

This logic is used in `packages/cli/src/cli.ts` during task input parsing:

https://github.com/TabbyML/pochi/blob/main/packages/cli/src/cli.ts#L292-L298

And `packages/cli/src/on-override-messages.ts` was also mentioned as relevant context for message processing.

**Suggested Fix**
The regex for slash commands should be more specific, possibly requiring a leading space or start of line, or explicitly excluding patterns that look like URLs (e.g. preceded by `://` or part of a larger non-whitespace string that doesn't start with `/`).

**Information about your version**
Current codebase.

🤖 Generated with [Pochi](https://getpochi.com)

PR #1026, opened 2026-01-05
fix(cli): improve slash command detection by parsing markdown AST

liangfung:
## Summary
- Use `remark` and `unist-util-visit` to parse markdown and extract slash commands only from text nodes.
- Ignore slash-like patterns within code blocks, inline code, links, images, and HTML tags.
- Add comprehensive tests for markdown edge cases in slash command matching.
- Update dependencies to include markdown processing utilities.

## Screenshot
image

## Test plan
- Run `bun run test` in `packages/cli` to verify the new tests pass.
- Manually verify that slash commands are still detected correctly in normal text.
- Verify that URLs and code blocks no longer trigger slash command detection.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -47,10 +47,16 @@
         "cli-table3": "^0.6.5",
         "commander": "^14.0.0",
         "listr2": "^9.0.3",
+        "mdast": "^3.0.0",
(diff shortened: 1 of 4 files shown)


liangfung:
fix #986

PR #1062, opened 2026-01-09
refactor: change sidebar task storage to use VSCode global state

wsxiaoys:
## Summary
Refactor task storage to use VSCode global state instead of client-side SQLite.

- Moves task storage logic to `packages/vscode/src/lib/task-store.ts` using `globalState`.
- Updates `vscode-webui` to read tasks from the host via `readTasks`.
- Removes `livestore` related task schema and queries from `packages/livekit` and `packages/vscode-webui`.
- Simplifies `vscode-webui` hooks for tasks and worktrees.

## Test plan
- Verify that tasks persist across VSCode reloads.
- Verify that the task list in the sidebar loads correctly.
- Verify that creating new tasks and subtasks works as expected.
- Verify that deleted worktree detection still works.

### DIFF
--- a/packages/vscode/src/lib/task-store.ts
+++ b/packages/vscode/src/lib/task-store.ts
@@ -1,6 +1,8 @@
+import { TextDecoder, TextEncoder } from "node:util";
 import { taskUpdated } from "@/lib/task-events";
 import { getLogger } from "@getpochi/common";
 import { signal } from "@preact/signals-core";
+import { funnel } from "remeda";
 import { inject, injectable, singleton } from "tsyringe";
 import * as vscode from "vscode";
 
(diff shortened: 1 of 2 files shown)


packages/vscode/src/lib/task-store.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  private loadTasks() {
+    const tasks = this.context.globalState.get<Record<string, EncodedTask>>(
+      this.storageKey,
+      {},
+    );
+
+    const now = Date.now();
+    const threeMonthsInMs = 90 * 24 * 60 * 60 * 1000;
+    const cutoff = now - threeMonthsInMs;
```
wsxiaoys, line 68: @zhanba fyi - stale task removing logic

PR #1064, opened 2026-01-09
fix: remove useToolCallMiddleware from vendor options

wsxiaoys:
## Summary
This PR removes the `useToolCallMiddleware` property from the model options in `packages/vendor-pochi/src/vendor.ts`. This property was conditionally set based on the model ID containing "google".

## Test plan
- Verify that the application builds and runs correctly.
- Ensure that models (especially those with "google" in the ID) still function as expected without this middleware option being explicitly set here.


wsxiaoys:
As gemini 3 series are much more stable on tool calling - we no longer need to use the customized tool call middleware

PR #1068, opened 2026-01-10
fix: add check for tabGroups length to avoid undefined access

wsxiaoys:
## Summary
- Added a check for `vscode.window.tabGroups.all.length > 0` before accessing the first element to prevent potential runtime errors when no tab groups exist.

## Test plan
- Verify that the extension no longer throws an error when `tabGroups.all` is empty.


wsxiaoys:
```
2026-01-10 13:21:21.258 [error] [PochiWebviewPanel] Failed to open Pochi task: Cannot read properties of undefined (reading 'tabs') {"nativeError":{},"name":"TypeError","message":"Cannot read properties of undefined (reading 'tabs')","stack":[]}
```

PR #1077, opened 2026-01-13
chore(vscode): update .vscodeignore to exclude test and env files

icycodes:
## Summary
- Updated `packages/vscode/.vscodeignore` to exclude development and test-related files.
- Added `.env`, `test/**`, `wdio.conf.ts`, and `.wdio-vscode-service/**` to the ignore list.
- Ensured `coverage/**` is correctly ignored.

## Test plan
- Verified that the changes are correctly staged and committed.
- Checked that the files listed are indeed development/test specific and should not be included in the extension package.

### DIFF
--- a/packages/vscode/.vscodeignore
+++ b/packages/vscode/.vscodeignore
@@ -11,6 +11,6 @@ turbo.json
 .turbo/**
 coverage/**
 .env
-test/**
+tests/**
 wdio.conf.ts
 .wdio-vscode-service/**


packages/vscode/.vscodeignore
```diff
@@ -9,4 +9,8 @@ tsconfig.test.json
 *.vsix
 turbo.json
 .turbo/**
-coverage/**
\ No newline at end of file
+coverage/**
+.env
+test/**
```
wsxiaoys:
```suggestion
tests/**
```

Please rebase against main (i just adjust integration test dir structure a bit)

icycodes: Updated

PR #1143, opened 2026-01-22
feat(vscode, cli): add agent skill support to cli and vscode extension

ziqch:
## Summary

This PR introduces agent skills support, enabling users to define custom agent skills in SKILL.md files and invoke them through slash commands (`/skill-name`) or automatic discovery via the `Skill tool`. Skills are automatically discovered from workspace directories and provide enhanced agent capabilities.

## Screenshots
CleanShot 2026-01-22 at 10 26 09@2x
CleanShot 2026-01-22 at 10 36 33@2x

### Core Skill System

- **Skill Definition Format**: Support for SKILL.md files with frontmatter (`packages/common/src/tool-utils/skill-parser.ts`)
- **Type Definitions**: Comprehensive TypeScript types for skills and validation (`packages/common/src/vscode-webui-bridge/types/skill.ts`)
- **Skill Tool**: Added Skill tool for automatic skill discovery and execution (`packages/cli/src/tools/skill.ts`, `packages/vscode/src/tools/skill.ts`)

### CLI Support

- **Skill Loading**: Automatic discovery from workspace directories (`packages/cli/src/lib/load-skills.ts`)
- **Slash Command Integration**: Enhanced command matching for skills (`packages/cli/src/lib/match-slash-command.ts`)
- **Task Runner Integration**: Updated to support skill execution (`packages/cli/src/task-runner.ts`)

### VSCode Extension Support

- **Skill Manager**: Workspace-scoped skill management with caching (`packages/vscode/src/lib/skill-manager.ts`)
- **Settings UI**: Skill management interface (`packages/vscode-webui/src/features/settings/components/sections/skill-section.tsx`)
- **Slash Command Autocomplete**: Skills in prompt suggestions (`packages/vscode-webui/src/components/prompt-form/slash-mention/`)
- **Webview Bridge**: Skills support in webview communication (`packages/common/src/vscode-webui-bridge/webview.ts`)

### User Experience

- **Auto-Discovery**: Skills from `.pochi/skills/` and `~/.pochi/skills/` directories
- **Error Handling**: Comprehensive validation with detailed error messages
- **Internationalization**: Added translations for EN/JP/KO/ZH (`packages/vscode-webui/src/i18n/locales/`)

## Test Plan

- [ ] **Skill Discovery**: Create skills in `.pochi/skills/test-skill/SKILL.md`, verify detection
- [ ] **Manual Invocation**: Test `/skill-name` commands in CLI and VSCode
- [ ] **Automatic Discovery**: Verify Skill tool finds and executes relevant skills
- [ ] **Slash Command Integration**: Check autocomplete suggestions and execution
- [ ] **Error Handling**: Test malformed files and invalid skill references
- [ ] **UI Integration**: Verify skill section in VSCode settings
- [ ] **Internationalization**: Test UI in different languages

## Files Changed

- **Core Infrastructure (16 files)**: Skill parsing, validation, and tool implementations
- **CLI Integration (7 files)**: Skill loading, command matching, and task runner updates
- **VSCode Extension (17 files)**: Skill manager, settings UI, webview bridge, and i18n

### DIFF
--- a/packages/cli/src/lib/__tests__/match-slash-command.test.ts
+++ b/packages/cli/src/lib/__tests__/match-slash-command.test.ts
@@ -189,7 +189,7 @@ describe("match-slash-command", () => {
         skills,
       });
       expect(result).toBe(
-        'Please use <skill id="test-skill" path=".pochi/skills/test-skill/SKILL.md">test-skill</skill> for this task',
+        'Please use <skill id="test-skill" path=".pochi/skills/test-skill/SKILL.md">useSkill:test-skill</skill> for this task',
       );
     });
 
@@ -201,7 +201,7 @@ describe("match-slash-command", () => {
         skills,
       });
       expect(result).toBe(
-        'Use <workflow id="test-workflow" path=".pochi/workflows/test-workflow.md">This is a test workflow</workflow> and then <custom-agent id="test-agent" path=".pochi/agents/test-agent.md">newTask:test-agent</custom-agent> and finally <skill id="test-skill" path=".pochi/skills/test-skill/SKILL.md">test-skill</skill>',
+        'Use <workflow id="test-workflow" path=".pochi/workflows/test-workflow.md">This is a test workflow</workflow> and then <custom-agent id="test-agent" path=".pochi/agents/test-agent.md">newTask:test-agent</custom-agent> and finally <skill id="test-skill" path=".pochi/skills/test-skill/SKILL.md">useSkill:test-skill</skill>',
       );
     });
   });


packages/tools/src/index.ts
```diff
@@ -100,6 +103,7 @@ const createCliTools = (options?: CreateToolOptions) => ({
   globFiles,
   listFiles,
   readFile: createReadFileTool(options?.contentType),
+  skill: createSkillTool(options?.skills),
```
zhanba:
```suggestion
  useSkill: createUseSkillTool(options?.skills),
```
tool name should start with verb

packages/tools/src/use-skill.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const inputSchema = z.object({
+  skill: z.string().describe("The name of the skill to use."),
+});
+
+export const createSkillTool = (skills?: Skill[]) => {
+  return defineClientTool({
+    description: `Execute a skill within the main conversation
+
```
zhanba, line 54: mention how to use optional directories like scripts/, references/, and assets/ ?

packages/tools/src/index.ts
zhanba: this tool should be enabled by default in subagent (see overrideCustomAgentTools)

ziqch: Fixed. Thanks for the suggestion!

packages/tools/src/use-skill.ts
ziqch: Updated the prompt to clarify how to use these optional directories.

packages/tools/src/index.ts
ziqch: Good catch! Enabled by default for subagents now. Thanks!

packages/cli/src/tools/use-skill.ts
wsxiaoys, line 1: rename to use-skill.ts. this also applies other tool related defs in tools / vscode packages

packages/vscode/src/tools/skill.ts
```diff
@@ -0,0 +1,54 @@
...
+        result: `Skill "${args.skill}" is invalid: ${
+          invalidSkill.message || "Unknown error"
+        }`,
+      };
+    }
+
+    logger.debug(`Activating skill: ${skill.name}`);
+
+    return {
+      result: skill.instructions.trim(),
```
wsxiaoys: instead of solely return skill's instruction, i believe you also need to construct a prompt based on allowed-tools to request model only use those tools.

packages/vscode/src/tools/use-skill.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // Check if skill is valid
+    if (!isValidSkillFile(skill)) {
+      const invalidSkill = skill as { message?: string };
+      return {
+        result: `Skill "${args.skill}" is invalid: ${
+          invalidSkill.message || "Unknown error"
+        }`,
+      };
+    }
```
wsxiaoys: SkillManager should only provide valid skills - thus i don't feel the check is necessary here

packages/tools/src/skill.ts
```diff
@@ -0,0 +1,90 @@
...
+function makeSkillToolDescription(skills?: Skill[]) {
+  if (!skills || skills.length === 0)
+    return "No skills are available in the workspace.";
+
+  return `Available skills:
+
+${skills
+  .map(
+    (skill) =>
+      `- **${skill.name}**: ${skill.description.trim()} [Location: ${skill.filePath}]`,
```
wsxiaoys: need to include compatibility here so llm got a chance to decide whether it's appropriate to use this skill based on os / environment etc

packages/cli/src/tools/use-skill.ts
ziqch: Fixed. Thanks for the suggestion!

packages/vscode/src/tools/skill.ts
ziqch: Good point! I’ve added a new helper `makeUseSkillResult` to construct the prompt, explicitly includes the allowed-tools so the model is constrained to only use those tools.

packages/vscode/src/tools/use-skill.ts
ziqch: You’re right! The skill name passed into `useSkill` are guaranteed to be valid, but `SkillManager` still needs to keep invalid skills so we can surface validation info in the skill section. I think this check helps ensure type correctness at the `useSkill`.

CleanShot 2026-01-24 at 19 16 26@2x

packages/tools/src/skill.ts
ziqch: I’ve updated the prompt, it now includes the compatibility information. Thanks!

packages/vscode/src/tools/use-skill.ts
wsxiaoys: in that case - we can provide allSkills / validSkills in SkillManager for different purpose (ui display / skill execution)

ziqch: Yeah, agreed, it is also a clean way to separate concerns. I’ve added allSkills / validSkills to SkillManager. Thanks

packages/vscode/src/lib/skill-manager.ts
```diff
@@ -0,0 +1,158 @@
...
+  return skills;
+}
+
+@scoped(Lifecycle.ContainerScoped)
+@injectable()
+export class SkillManager implements vscode.Disposable {
+  private disposables: vscode.Disposable[] = [];
+
+  readonly allSkills = signal<SkillFile[]>([]);
+  readonly validSkills = signal<ValidSkillFile[]>([]);
```
wsxiaoys:
```suggestion
  readonly skills = signal<SkillFile[]>([]);
  readonly validSkills = computed(...)
```

packages/tools/src/use-skill.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export const inputSchema = z.object({
+  skill: z.string().describe("The name of the skill to use."),
+});
+
+/**
+ * Creates the result for useSkill tool that includes skill instructions and tool restrictions
+ */
+export function makeUseSkillResult(skill: Skill): string {
```
wsxiaoys: move to package common/prompts

packages/vscode/src/lib/skill-manager.ts
ziqch: Using a computed here makes more sense, thanks for pointing it out.

packages/tools/src/use-skill.ts
ziqch: Moved it to the common package.

PR #1154, opened 2026-01-22
feat: add attempt-completion-schema and hook

DESU-CLUB:
## Summary
This PR introduces the `attempt-completion-schema` tool and `attempt-completion-hook` mechanism in the task runner. This introduces two new flags for validating and running commands on attempt-completion.

- **`attempt-completion-schema` tool**: A new tool that specifies a schema for attempt-completion and provide a final result, in that exact schema.
- **`attempt-completion-hook`**: Added support for an `attemptCompletionHook` in the `TaskRunner`. When the `attemptCompletion` tool is called, the runner can execute a specified shell command to verify the result. The command receives the tool's input via stdin.
- **CLI Enhancements**: Updated the CLI to support the new `attempt-completion` flags and provide better output rendering for `attempt-completion-hook` results.

## What was changed
- `--attempt-completion-schema`:
  - This new flag allows us to specify a Zod schema which is enforced on attemptCompletion. This helps us create structured outputs on attemptCompletion tool calls, which can then be processed later using `attempt-completion-hook`.
  - Previously, the flow of attemptCompletion is as follows: 
    -  `task-runner.ts` would create a `Livekit<Chat>`, which would use `flexible-chat-transport.ts`, which then calls `CreateCLITools` which runs `attemptCompletion` tool. The schema for attemptCompletion was previously fixed to a specific schema: `{result" ....}`
    - There is no change to the workflow, but `cli.ts` now supplies a new flag via `attempt-completion-schema`, which propagates a schema along the same flow, and a new tool has been implemented in `attempt-completion.ts` to override the previous schema.
  - However, the renderer has been changed, as it used to expect the fixed schema from `attemptCompletion`. Now a boolean is passed into the renderer as well, to verify if the schema has been overriden so that it can render the new schema properly.

- `--attempt-completion-hook`
  - This flag provides us with a flexible command which we can run on the output of `attemptCompletion`
    - This is done by spawning a new command shell using `spawn`, and executing the command, using the stdin which outputs content from `attempt-completion` tool.
    -  For example, a tool like `bun run process_output.ts` can be run on the output to process the content from `attempt-completion`. This allows for introducing workflows that require bash/programmatic scripts to run on the output of Pochi

## Test plan
- [x] Verified that `attemptCompletion` tool is correctly defined and exported.
- [x] Verified that `TaskRunner` correctly handles the `attemptCompletionHook` and executes the verification command.
- [x] Tested the CLI output rendering for the new tool.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -255,19 +255,19 @@ const program = new Command()
     });
 
     const renderer = new OutputRenderer(runner.state, {
-      hasCustomAttemptCompletionSchema: !!options.attemptCompletionSchema,
+      attemptCompletionSchemaOverride: !!options.attemptCompletionSchema,
     });
 
     let jsonRenderer: JsonRenderer | undefined;
     if (options.streamJson) {
       jsonRenderer = new JsonRenderer(blobStore, runner.state, {
         mode: "full",
-        hasCustomAttemptCompletionSchema: !!options.attemptCompletionSchema,
+        attemptCompletionSchemaOverride: !!options.attemptCompletionSchema,
       });
     } else if (options.outputResult) {
       jsonRenderer = new JsonRenderer(blobStore, runner.state, {
         mode: "result-only",
-        hasCustomAttemptCompletionSchema: !!options.attemptCompletionSchema,
+        attemptCompletionSchemaOverride: !!options.attemptCompletionSchema,
       });
     }
 
--- a/packages/cli/src/task-runner.ts
+++ b/packages/cli/src/task-runner.ts
@@ -124,7 +124,7 @@ export class TaskRunner {
 
   readonly taskId: string;
 
-  readonly hasCustomAttemptCompletionSchema: boolean;
+  readonly attemptCompletionSchemaOverride: boolean;
 
   private get chat() {
     return this.chatKit.chat;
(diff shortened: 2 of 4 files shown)


packages/cli/src/cli.ts
```diff
@@ -233,17 +248,26 @@ const program = new Command()
...
         ? parseOutputSchema(options.experimentalOutputSchema)
         : undefined,
+      attemptCompletionSchema: options.attemptCompletionSchema
+        ? parseOutputSchema(options.attemptCompletionSchema)
+        : undefined,
+      attemptCompletionHook: options.attemptCompletionHook,
+    });
+
+    const renderer = new OutputRenderer(runner.state, {
+      hasCustomAttemptCompletionSchema: !!options.attemptCompletionSchema,
```
wsxiaoys:
```suggestion
      attemptCompletionSchemaOverride: !!options.attemptCompletionSchema,
```

nit, please also update other places

packages/cli/src/task-runner.ts
```diff
@@ -384,6 +420,54 @@ export class TaskRunner {
 
     return "next" as const;
   }
+
+  // Helper method to run the command
+  private runVerificationCommand(
```
wsxiaoys:
```suggestion
  private runAttemptCompletionHook(
```

PR #1165, opened 2026-01-24
fix: fix background job output rendering and localize async label

leslieo2:
## Summary
  - replace hardcoded "ASYNC" label with i18n key and add translations
  - make XTerm render \n as real line breaks to fix background job output display
  
  before:
image
fixed:
image

### DIFF
--- a/packages/vscode-webui/src/features/tools/components/new-task/index.tsx
+++ b/packages/vscode-webui/src/features/tools/components/new-task/index.tsx
@@ -91,7 +91,7 @@ function BackgroundTaskToolView(
           onClick={openInTab}
           disabled={!canOpen}
         >
-          ASYNC
+          {"ASYNC"}
         </Button>
       </ToolTitle>
     </div>
--- a/packages/vscode-webui/src/features/tools/components/xterm.tsx
+++ b/packages/vscode-webui/src/features/tools/components/xterm.tsx
@@ -240,6 +240,7 @@ export function XTerm({
   ...props
 }: XTermProps) {
   const defaultTerminalOptions = {
+    convertEol: true,
     rows: minRow,
   };
 


packages/vscode-webui/src/features/tools/components/new-task/index.tsx
```diff
@@ -91,7 +93,7 @@ function BackgroundTaskToolView(
           onClick={openInTab}
           disabled={!canOpen}
         >
-          ASYNC
```
wsxiaoys, line 94: Please revert

leslieo2: just to pass bun run check

leslieo2: 93:10  error  disallow literal string: <Button
          variant="ghost"
          size="sm"
          className="h-6 px-2 text-xs"
          onClick={openInTab}
          disabled={!canOpen}
        >
          ASYNC
        </Button>  i18next/no-literal-string

✖ 1 problem (1 error, 0 warnings)

error: script "check:i18n" exited with code 1
error: script "check:i18n" exited with code 1
ERROR: "check:i18n" exited with 1.
error: script "check" exited with code 1

wsxiaoys: please add an lint ignore for ths line

leslieo2: done

wsxiaoys:
Could you also provide screenshot / story book for readBackgroundJob of a terminal job to ensure this updated config doesn't break that?

leslieo2:
> Could you also provide screenshot / story book for readBackgroundJob of a terminal job to ensure this updated config doesn't break that?
after
image

PR #1192, opened 2026-01-28
feat(cli): support planner subagent in pochi cli

wsxiaoys:
## Summary
- Refactored `resolveToolCallArgs` and `resolvePochiUri` utilities to `@getpochi/common` for sharing between CLI and VSCode.
- Implemented a `FileSystem` abstraction in CLI to support Virtual File System (VFS) using `pochi://` URIs.
- Updated CLI tools (`applyDiff`, `editNotebook`, `readFile`, `writeToFile`) to use the new `FileSystem` interface.
- Enhanced `resolveToolCallArgs` to correctly handle nested objects and arrays in tool call arguments.

## Test plan
- Verified that CLI tools correctly handle both local files and `pochi://` URIs.
- Added unit tests for `resolveToolCallArgs` in `packages/common`.
- Ran CLI integration tests to ensure no regressions in tool execution.

### DIFF
--- a/packages/cli/src/lib/file-system.ts
+++ b/packages/cli/src/lib/file-system.ts
@@ -50,6 +50,12 @@ export class TaskFileSystem implements FileSystem {
       throw new Error(`Invalid task URI: ${filePath}`);
     }
 
+    if (uri.filePath !== "/plan.md") {
+      throw new Error(
+        `Only /plan.md is supported for task file system, got: ${uri.filePath}`,
+      );
+    }
+
     await this.store.commit(
       catalog.events.writeTaskFile({
         taskId: uri.taskId,


packages/vscode/src/integrations/webview/vscode-host-impl.ts
```diff
@@ -534,26 +537,16 @@ export class VSCodeHostImpl implements VSCodeHostApi, vscode.Disposable {
...
-    if (fileUri.scheme === "pochi" && this.task) {
-      resolvedPath = resolvePochiUri(filePath, this.task.id);
-      vscode.commands.executeCommand(
-        "vscode.open",
-        vscode.Uri.parse(resolvedPath),
-      );
-      return;
-    }
+    if (!this.task) return;
+    let resolvedPath = resolvePochiUri(filePath, this.task.id);
```
godtail: The Pochi schema file becomes inaccessible (throwing a "file not found" error), possibly due to interference from later file operations.

packages/cli/src/task-runner.ts
```diff
@@ -158,7 +175,6 @@ export class TaskRunner {
 
         const runner = new TaskRunner({
           ...options,
-          blobStore: this.blobStore,
```
godtail, line 161: Is it okay to remove this option? It seems unrelated to this PR

wsxiaoys: it's filled from `…options` already

packages/vscode/src/integrations/webview/vscode-host-impl.ts
wsxiaoys: fixed

PR #1205, opened 2026-01-29
feat(cli): Support attaching URLs in Pochi CLI

DESU-CLUB:
### Summary
- This provides a partial reference for how to handle URLs in Pochi
- Added partial handling for URLs in `cli.ts`. This allows us to now be able to attach urls and fetch the information from URLs in CLI.
- With current implementation, it will also not OOM on large video files

### Unsolved Bugs/Blockers:
- When videos and text based URL are attached, Pochi currently hang on user prompt, and Pochi times out.
- I have tested that at least for text based URL, this is due to enabling all URLs in `vendor-pochi/model.ts`, under the variable `supportedURLs` under the function `createPochiModel`
- This variable changes the behaviour of Pochi, which forces it to attach the URL as a message, rather than download the file and sending the data of the file inline, which is the main cause of OOM for large video files.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -24,10 +24,7 @@ import {
 } from "@getpochi/common/configuration";
 import { getVendor, getVendors } from "@getpochi/common/vendor";
 import { createModel } from "@getpochi/common/vendor/edge";
-import {
-  type LLMRequestData,
-  type Message,
-} from "@getpochi/livekit";
+import type { LLMRequestData, Message } from "@getpochi/livekit";
 import chalk from "chalk";
 import * as commander from "commander";
 import z from "zod/v4";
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli.ts
```diff
@@ -176,25 +176,72 @@ const program = new Command()
...
-          const dataUrl = await fileToUri(
-            blobStore,
-            new File([buffer], attachmentPath, {
-              type: mimeType,
-            }),
-          );
+          const isUrl =
+            attachmentPath.trim().startsWith("http://") ||
+            attachmentPath.trim().startsWith("https://") ||
+            attachmentPath.trim().startsWith("gs://);
```
wsxiaoys: no need to check for protocol, as long as it's a url (not local path, let's just pass it through to live kit).

livekit will do the check internally

packages/cli/src/cli.ts
```diff
@@ -176,25 +176,72 @@ const program = new Command()
     if (attachments && attachments.length > 0) {
```
wsxiaoys: i aslo recommend extract this to attachment-utils to reduce size of this file

wsxiaoys (changes requested):

PR #1228, opened 2026-02-03
fix(cli): attempt-completion hook does not work during streaming completions

DESU-CLUB:
### Summary
- Add fix for completion hook, and improve streaming UI
- Instead of streaming multiple incomplete JSONs, we just display that the answer is still streaming. A sample is attached below.
- Moreover, there was a previous issue with completion hook, where it would just give an error. The main cause is due to `output-renderer.ts` not catching all possible states, and sending some valid requests to the error branch of the if conditional (line 90 of `output-renderer.ts`)
  - This has been fixed by updating the if condition to also accept "call" in the state.
Screenshot 2026-02-02 at 9 15 14 PM

### DIFF
--- a/packages/cli/src/output-renderer.ts
+++ b/packages/cli/src/output-renderer.ts
@@ -263,12 +263,12 @@ function renderToolPart(
     const input = part.input || {};
 
     let content = "";
-    if (part.state === "input-streaming"){
+    if (part.state === "input-streaming") {
       return {
         text: "",
         stop: "stopAndPersist",
         error: errorText,
-      }
+      };
     }
     if (attemptCompletionSchemaOverride) {
       content = JSON.stringify(input.result, null, 2);
--- a/packages/tools/src/attempt-completion.ts
+++ b/packages/tools/src/attempt-completion.ts
@@ -8,7 +8,7 @@ export const attemptCompletionSchema = z.object({
     .describe(
       "The result of the task. Formulate this result in a way that is final and does not require further input from the user.",
     ),
-  });
+});
 
 const toolDef = {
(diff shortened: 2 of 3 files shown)


packages/cli/src/output-renderer.ts
```diff
@@ -83,10 +83,14 @@ export class OutputRenderer {
...
 
+        // @ts-ignore 
+        // We need to check for "call" as well, or else hook will not receive data
+        const state = part.state;
         if (
-          (isAutoSuccessToolPart(part) && part.state === "input-available") ||
-          part.state === "output-available" ||
-          part.state === "output-error"
+          (isAutoSuccessToolPart(part) &&
+            (state === "input-available" || state === "call")) ||
```
wsxiaoys:
```suggestion
            (state === "input-available")) ||
```

packages/tools/src/attempt-completion.ts
```diff
@@ -29,9 +31,26 @@ ${NoOtherToolsReminderPrompt}
...
 export const attemptCompletion = defineClientTool(toolDef);
 
-export const createAttemptCompletionTool = (schema?: z.ZodAny) =>
-  defineClientTool({
+export const createAttemptCompletionTool = (schema?: z.ZodAny) => {
+  if (!schema) {
+    return defineClientTool(toolDef);
+  }
+
+  // Preprocess to handle cases where the model returns a JSON string instead of an object
```
wsxiaoys: no need to do this - model api already gurantee it shall be an json object. and if it's not - it will be handled by repair-tool-call anyway

PR #1232, opened 2026-02-03
fix(vscode-webui): use custom navigation hook to support share page

liangfung:
## Summary
This PR fixes a bug where `useNavigate` would cause an error in the share page (where no router context exists). Instead of adding a dummy router, we now use a custom `useVSCodeNavigate` hook that returns a no-op function when not running in the VSCode environment.

## Test plan
- Verify that the share page loads correctly without errors.
- Verify that navigation works as expected within VSCode.
- Verify that components using `useVSCodeNavigate` do not crash in the share page.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-cdcaac4eb9fe40f0932063ec4bc57fa0)

### DIFF
--- a/packages/vscode-webui/src/lib/hooks/use-navigate.ts
+++ b/packages/vscode-webui/src/lib/hooks/use-navigate.ts
@@ -1,7 +1,7 @@
-import { useNavigate as useTanStackNavigate } from "@tanstack/react-router";
+import { useNavigate as useTanstackNavigate } from "@tanstack/react-router";
 import { isVSCodeEnvironment } from "../vscode";
 
-type NavigateFn = ReturnType<typeof useTanStackNavigate>;
+type NavigateFn = ReturnType<typeof useTanstackNavigate>;
 
 const noopNavigate = (() => {}) as unknown as NavigateFn;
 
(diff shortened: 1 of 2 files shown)


packages/vscode-webui/src/lib/hooks/use-vscode-navigate.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const noopNavigate = (() => {}) as unknown as NavigateFn;
+
+/**
+ * Navigate hook that only works in VSCode webview environment.
+ *
+ * In VSCode webview: returns the real useNavigate() from TanStack Router.
+ * In other environments (e.g., share page iframe): returns a no-op function.
+ */
+export function useVSCodeNavigate(): NavigateFn {
```
wsxiaoys:
```suggestion
export function useNavigate(): NavigateFn {
```

just call this useNavigate, and disable import of tanstack router useNavigate (except for this file).

see rules/no-console.yaml for example

liangfung: updated

rules/no-tanstack-use-navigate.yaml
wsxiaoys, line 1: no-tanstack-use-navigate

packages/vscode-webui/src/lib/hooks/use-navigate.ts
```diff
@@ -0,0 +1,17 @@
+import { useNavigate as useTanStackNavigate } from "@tanstack/react-router";
```
wsxiaoys:
```suggestion
import { useNavigate as useTanstackNavigate } from "@tanstack/react-router";
```

PR #1263, opened 2026-02-09
fix(vscode): improve handling of 402/429 errors in status bar

liangfung:
## Summary
- Detect 402 (Payment Required) and 429 (Rate Limit Exceeded) errors in completion manager.
- Expose error state in completion manager.
- Update status bar to show warning icon and error message tooltip when these errors occur.

## Test plan
- Manually verified that 402/429 errors trigger the warning icon in the status bar.
- Verified that other errors do not trigger the warning icon.
- Verified that the error clears when the condition is resolved.

## Screenshot
image

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-76d49ac7dacc4df9bd60a6658432cb2f)

### DIFF
--- a/packages/vscode/src/integrations/status-bar-item.ts
+++ b/packages/vscode/src/integrations/status-bar-item.ts
@@ -185,7 +185,11 @@ export class StatusBarItem implements vscode.Disposable {
         this.statusBarItem.text = "$(warning) Pochi";
         this.statusBarItem.tooltip = this.tabCompletionManager.error.value;
         this.statusBarItem.backgroundColor = undefined;
-        this.statusBarItem.command = undefined;
+        this.statusBarItem.command = {
+          command: "pochi.showWarningMessage",
+          title: "",
+          arguments: [this.tabCompletionManager.error.value],
+        };
         break;
 
       case "ready":
--- a/packages/vscode/src/tab-completion/completion-manager.ts
+++ b/packages/vscode/src/tab-completion/completion-manager.ts
@@ -36,8 +36,6 @@ import {
 import {
   delayFn,
   isLineEndPosition,
-  isPaymentRequiredError,
-  isRateLimitExceededError,
   offsetRangeToPositionRange,
 } from "./utils";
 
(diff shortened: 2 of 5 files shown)


packages/vscode/src/tab-completion/completion-manager.ts
```diff
@@ -468,6 +424,19 @@ export class TabCompletionManager implements vscode.Disposable {
...
+            if (
+              isRateLimitExceededError(status.error) ||
+              isPaymentRequiredError(status.error)
+            ) {
+              this.updateError(status.error.message);
+            } else {
+              // ignore
+              this.updateError(undefined);
+            }
+          }
```
icycodes: Updating error info directly in CompletionManager can lead to race conditions when multiple providers send requests at the same time, causing error messages to be lost. 
We should adding an error message signal to each provider. The CompletionManager can then combine these error message signals from all providers.

packages/vscode/src/tab-completion/completion-manager.ts
```diff
@@ -636,9 +629,16 @@ class TabCompletionManagerContext implements vscode.Disposable {
...
+}
+
+function combineProviderErrors(
+  providers: TabCompletionProvider[],
+): string | undefined {
+  for (const provider of providers) {
+    if (provider.error.value) {
+      return provider.error.value;
     }
   }
```
icycodes:
```suggestion
  const error = providers.map((provider) => provider.error.value ? `${provider.id}: ${provider.error.value}` : undefined).filter(Boolean).join("\n")
  return error || undefined;
```

packages/vscode/src/tab-completion/providers/provider.ts
```diff
@@ -26,4 +30,16 @@ export class TabCompletionProvider {
...
     );
   }
+
+  updateError(error: string | undefined) {
+    this.error.value = error;
+  }
+
+  clearError() {
+    this.error.value = undefined;
+  }
```
icycodes: Remove these functions, subscribe request status in provider itself.

packages/vscode/src/integrations/status-bar-item.ts
```diff
@@ -171,6 +181,13 @@ export class StatusBarItem implements vscode.Disposable {
         this.statusBarItem.command = undefined;
         break;
 
+      case "error":
+        this.statusBarItem.text = "$(warning) Pochi";
+        this.statusBarItem.tooltip = this.tabCompletionManager.error.value;
+        this.statusBarItem.backgroundColor = undefined;
+        this.statusBarItem.command = undefined;
```
icycodes: Show a warning message notification.

icycodes (approved):

PR #1264, opened 2026-02-09
fix(vscode): make VSCodeHostImpl stateless by passing taskId explicitly

wsxiaoys:
## Summary
This PR refactors `VSCodeHostImpl` to be stateless regarding the current task. It removes the `currentTaskId` property and the `taskId` setter, and instead requires `taskId` to be passed explicitly to methods that depend on it (`executeToolCall`, `previewToolCall`, `openFile`, `readEnvironment`).

It also updates `vscode-webui` to pass `taskId` where necessary.

Key changes:
- `VSCodeHostImpl` methods now accept `taskId` in options.
- `readEnvironment` now takes `taskId` and looks up `shareId` from `taskHistoryStore`.
- `useLiveChatKitGetters` and `PublicShareButton` updated to pass `taskId`.
- Added comments to `VSCodeHostApi` methods explaining that `taskId` is always the top-level task ID.

## Test plan
- Verify that tool calls still work correctly in the chat.
- Verify that tool previews work.
- Verify that opening files from the chat works.
- Verify that the share button works.
- Verify that `readEnvironment` correctly returns environment info with `shareId`.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-cbc1744f9bf54546b3fac42c37651dcd)

### DIFF
--- a/packages/vscode/src/integrations/webview/vscode-host-impl.ts
+++ b/packages/vscode/src/integrations/webview/vscode-host-impl.ts
@@ -466,17 +466,10 @@ export class VSCodeHostImpl implements VSCodeHostApi, vscode.Disposable {
         };
       }
 
-      const task = this.taskHistoryStore.tasks.value[options.taskId];
-      if (!task) {
-        return {
-          error: "No task found.",
-        };
-      }
-
       const abortSignal = new ThreadAbortSignal(options.abortSignal);
       const envs = resolveToolCallEnvs(toolName, options.builtinSubAgentInfo);
       const toolCallStart = Date.now();
-      const resolvedArgs = resolveToolCallArgs(args, task.id);
+      const resolvedArgs = resolveToolCallArgs(args, options.taskId);
       const result = await safeCall(
         tool(resolvedArgs, {
           abortSignal,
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/lib/vscode.ts
```diff
@@ -236,7 +236,18 @@ function createVSCodeHost(): VSCodeHostApi {
...
   );
 
-  return thread.imports;
+  const vscodeHostApi: VSCodeHostApi = thread.imports;
+  const openFile: VSCodeHostApi["openFile"] = async (filePath, options) => {
+    return vscodeHostApi.openFile(filePath, {
+      ...options,
+      taskId: POCHI_TASK_INFO?.uid,
+    });
+  };
```
wsxiaoys, line 245: @godtail fyi - inject taskId to openFile args

wsxiaoys:
todo: pass taskId in readEnvironment to avoid useQuery's delay

packages/vscode/src/integrations/webview/vscode-host-impl.ts
```diff
@@ -568,15 +575,24 @@ export class VSCodeHostImpl implements VSCodeHostApi, vscode.Disposable {
...
     let fileUri = vscode.Uri.parse(filePath);
     let resolvedPath = filePath;
 
     // Open file directly if it's a pochi scheme
     if (fileUri.scheme === "pochi") {
-      resolvedPath = resolvePochiUri(filePath, this.task.id);
+      if (!options?.taskId) {
+        return;
+      }
+      const task = this.taskHistoryStore.tasks.value[options.taskId];
```
godtail: Since we only need the taskId, is it fine to just use options.taskId?

wsxiaoys: fixed

PR #1292, opened 2026-02-14
fix(vscode-webui): decouple frame subscription from browser session instance

godtail:
## Summary
- Move frame subscriptions to `BrowserSessionManager` to allow subscribing before session start.
- Ensure subscriptions persist even if `BrowserRecordingSession` is recreated.

## Test plan
- Verify that browser view streaming works correctly even when the session is re-established.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-8505f37eb23b4ae2bd7f801bd5350aa1)

### DIFF
--- a/packages/vscode-webui/src/lib/browser-session-manager.ts
+++ b/packages/vscode-webui/src/lib/browser-session-manager.ts
@@ -9,6 +9,8 @@ import { vscodeHost } from "./vscode";
 
 const logger = getLogger("BrowserRecordingManager");
 
+const frameSubscriptions = new Map<string, Set<(frame: string) => void>>();
+
 export class BrowserRecordingSession {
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/lib/browser-session-manager.ts
```diff
@@ -199,11 +197,17 @@ export class BrowserSessionManager {
   }
 
   subscribeFrame(taskId: string, callback: (frame: string) => void) {
-    const recordingSession = this.recordingSessions.get(taskId);
-    if (!recordingSession) {
-      return () => {};
+    if (!this.frameSubscriptions.has(taskId)) {
+      this.frameSubscriptions.set(taskId, new Set());
```
zhanba: delete frameSubscriptions in unregisterSession?

PR #1295, opened 2026-02-23
fix(vscode-webui): add white screen check before recording

godtail:
## Summary
This PR adds a check to ensure that browser recordings do not start with a blank white screen. It introduces an `isWhiteScreen` helper function that inspects the first frame of the recording. If the frame is predominantly white, the recording initialization is deferred until a non-white frame is received.

image

## Test plan
1. Start a browser session in VSCode.
2. Verify that the recording starts only when actual content is visible, avoiding initial white flashes.
3. Verify that the live view continues to update immediately.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-a449a89b5e804eff8f7d238748aeeb71)

### DIFF
--- a/packages/vscode-webui/src/lib/browser-session-manager.ts
+++ b/packages/vscode-webui/src/lib/browser-session-manager.ts
@@ -8,8 +8,10 @@ import type { useDefaultStore } from "./use-default-store";
 import { vscodeHost } from "./vscode";
 
 const logger = getLogger("BrowserRecordingManager");
+
 const frameSubscriptions = new Map<string, Set<(frame: string) => void>>();
-const WHITE_SCREEN_CHECK_INTERVAL = 500;
+const whiteScreenCheckInterval = 500;
+const websocketRetryInterval = 2500;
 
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/lib/browser-session-manager.ts
```diff
@@ -8,13 +8,49 @@ import type { useDefaultStore } from "./use-default-store";
 import { vscodeHost } from "./vscode";
 
 const logger = getLogger("BrowserRecordingManager");
-
 const frameSubscriptions = new Map<string, Set<(frame: string) => void>>();
+const WHITE_SCREEN_CHECK_INTERVAL = 500;
```
zhanba:
```suggestion
const WhiteScreenCheckInterval = 500;
```

godtail: Should we use `whiteScreenCheckInterval` instead of `WhiteScreenCheckInterval`, since we don't need to export it?

PR #1382, opened 2026-03-19
fix(common): keep empty reasoning parts with providerMetadata

zhanba:
## Summary

- `removeEmptyTextParts` previously filtered out all reasoning parts with empty text
- Reasoning parts that carry `providerMetadata` (e.g. OpenAI `itemId`) must be retained even when their text is empty, because they are referenced as `item_reference` in subsequent API calls
- Added two new test cases covering both the removal of empty reasoning parts without metadata and the retention of empty reasoning parts with `providerMetadata`

## Test plan

- [ ] Run `bun run test` in `packages/common` to verify the new test cases pass
- [ ] Test with OpenAI models that use reasoning to confirm multi-turn conversations work correctly after this fix

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-0fc29804f1c943868d621e3dea6db595)


zwpaper:
LGTM

PR #1434, opened 2026-04-03
feat: support omitting custom rules for agents

liangfung:
## Summary
- Introduced `omitPochiMd` property for `CustomAgent` to allow opting out of workspace-specific custom rules (e.g., `README.pochi.md`).
- Enabled this flag for `planner` and `explore` agents to reduce prompt noise and focus on their specific roles.
- Refactored `readEnvironment` to use `omitCustomRules` instead of `isSubTask` for determining when to include custom rules.
- Added support for `omitPochiMd` in agent file frontmatter.

## Test plan
- Added unit tests in `packages/common/src/base/prompts/__tests__/prompt.test.ts` to verify that custom rules are omitted when the flag is set.
- Added unit tests in `packages/common/src/tool-utils/__tests__/agent-parser.test.ts` to verify that the flag is correctly parsed from agent frontmatter.
- Verified that `planner` and `explore` agents now have `omitPochiMd: true`.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-710a0e142c0b4559a08f8a12ebda6913)

### DIFF
--- a/packages/common/src/base/agents/explore.ts
+++ b/packages/common/src/base/agents/explore.ts
@@ -2,7 +2,7 @@ import type { CustomAgent } from "@getpochi/tools";
 
 export const explore: CustomAgent = {
   name: "explore",
-  omitAgentMd: true,
+  omitAgentsMd: true,
   description: `
 Use this agent to explore the codebase when you need to: understand project structure, find where features are implemented, locate specific functions or classes, analyze code patterns, investigate how something works, search for examples or usage, or gather information before making changes.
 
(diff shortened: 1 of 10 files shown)


packages/common/src/base/agents/explore.ts
```diff
@@ -2,6 +2,7 @@ import type { CustomAgent } from "@getpochi/tools";
 
 export const explore: CustomAgent = {
   name: "explore",
+  omitAgentMd: true,
```
zhanba:
```suggestion
  omitAgentsMd: true,
```

PR #1474, opened 2026-04-20
fix(livestore): properly flatten decoded arrays in do-rpc client

godtail:
This patch simplifies the array unwrapping logic in `@livestore/common-cf`'s `do-rpc/client.js` by using `.flat(1)` instead of checking for specific nested arrays of length 1. This prevents issues with incorrectly nested or unexpectedly formatted response arrays.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-0927ed6d727a41528684412c0aab4673)


godtail:
Related PR: https://github.com/livestorejs/livestore/pull/1167

PR #1477, opened 2026-04-20
feat: refactor tool call parallel execution to micro-batching

liangfung:
## Summary
This PR refactors the tool call parallel execution mechanism into a more robust micro-batching system. Instead of simple parallel execution, tool calls are now partitioned into batches based on their side-effect profile:
- **Concurrent Batches**: Consecutive safe-to-batch (read-only or fire-and-forget) tool calls are executed concurrently.
- **Serial Batches**: Stateful or side-effecting tool calls (e.g., `executeCommand` with write operations, `newTask` without `runAsync`) act as serial barriers, ensuring they run in isolation and in order.

### Key Changes
- **`@getpochi/tools`**: Added `batch-utils.ts` for partitioning tool calls and managing batch execution with concurrency limits and error handling.
- **`vscode-webui`**: Introduced `BatchExecuteManager` to handle task-specific FIFO queues for tool calls, allowing sub-tasks and the main task to manage their own execution flow while sharing the same infrastructure.
- **Side-effect Analysis**: Implemented a conservative shell command parser to identify read-only `executeCommand` calls (e.g., `git log`, `grep`, `ls`) that are safe to run concurrently.
- **Error Handling**: Failures in a batch now correctly cancel subsequent pending tool calls for that specific task.

## Test plan
1. Verified that multiple `readFile` or `listFiles` calls are batched and executed concurrently in the UI.
2. Verified that a stateful command like `writeFile` acts as a barrier, forcing previous reads to complete before it starts, and subsequent calls to wait.
3. Tested error scenarios where a failed tool call in a batch correctly halts the remaining scheduled calls for that task.
4. Ran existing tests in `packages/vscode/src/tools/__test__/execute-command.test.ts` to ensure no regressions in command execution logic.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-67c6717186f54b6ab6f94e1b071b1ae7)


packages/tools/src/batch-utils.ts
```diff
@@ -0,0 +1,783 @@
...
+    batches.push({
+      mode: "concurrent",
+      items: currentConcurrentBatch,
+    });
+  }
+
+  return batches;
+}
+
+async function runConcurrentBatch<T>(
```
zhanba: add test?

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+/** Maximum concurrent executions within one concurrent batch. */
+const MaxConcurrency = 10;
+
+type ToolCallQueueOptions = {
+  getCustomAgents?: () => CustomAgent[] | undefined;
+  concurrencyLimit?: number;
+};
+
+/** FIFO queue for one `taskId`. */
+class ToolCallQueue {
```
zhanba: add test?

packages/tools/src/batch-utils.ts
```diff
@@ -0,0 +1,733 @@
...
+    execute: (
+      item: T,
+      batchMode: ToolBatchMode,
+      abortSignal: AbortSignal,
+    ) => Promise<void>;
+  },
+): Promise<void> {
+  const { concurrencyLimit, execute } = options;
+  const abortController = new AbortController();
+
```
icycodes: 1. `execute` should be a required top-level parameter, not nested in `options`.

2. `executePartitionedToolCalls` doesn’t accept an `AbortSignal` from the caller, yet it creates an internal `AbortController` and threads its signal into the callback. 
   I don’t think creating an `AbortController` is needed here.
   - Within a batch, we should just wait for all tool calls to finish, no need to “abort”.
   - Between batches, if a prior batch fails, we simply shouldn’t start the next batch, no need to abort.

packages/tools/src/batch-utils.ts
```diff
@@ -0,0 +1,733 @@
...
+        );
+      } else {
+        for (let itemIndex = 0; itemIndex < batch.items.length; itemIndex++) {
+          const item = batch.items[itemIndex];
+          if (!item) continue;
+
+          try {
+            await execute(item, batch.mode, abortController.signal);
+          } catch (error) {
+            abortController.abort(error);
```
icycodes: The usage of `abortController.abort(error)` here is not correct. We cannot `abort` an already failed `execute`.

packages/tools/src/batch-utils.ts
```diff
@@ -0,0 +1,733 @@
...
+import { ToolsByPermission } from "./constants";
+import type { CustomAgent } from "./new-task";
+import { getToolArgs, parseToolSpec } from "./utils";
+
+export type ToolBatchMode = "concurrent" | "serial";
+
+export type ToolBatch<T> = {
+  mode: ToolBatchMode;
+  items: T[];
+};
```
icycodes: I think we should remove `ToolBatchMode` from these type definitions.

For a list of tool calls, we can split it into a `T[][]`, where each batch is just a `T[]`. If a tool call is not read-only, it should be a single-element array (i.e., its own batch).

I also see some cancellation logic that depends on this `mode` flag, but I don’t think it’s required either. If a batch has at least one error, we should stop and not start the remaining batches, and return a `cancelled` state for them.

packages/tools/src/batch-utils.ts
```diff
@@ -0,0 +1,733 @@
...
+
+/**
+ * Returns `true` if the tool call has no side effects and can be executed
+ * concurrently with other read-only tool calls in the same microbatch.
+ */
+export function isReadonlyToolCall(
+  toolName: string,
+  input: unknown,
+  customAgents?: CustomAgent[],
+): boolean {
```
icycodes: I suggest splitting the files: move the read-only check into a separate file, and keep the batching logic in its own file.

packages/tools/src/batch-utils.ts
liangfung: The current design is that remaining tool calls will be canceled only when a non-readonly tool call encounters an error. If concurrent tool calls encounter an error, they will not cancel the remaining tool calls. 

Therefore, it is more appropriate to determine whether a batch is concurrency-safe based on the batch's mode. Judging solely by the number of elements in the batch is not sufficient.

packages/vscode/src/integrations/terminal/utils.ts
```diff
@@ -60,20 +60,6 @@ export async function* createBackgroundOutputStream(
   }
 }
 
-/**
```
zhanba, line 63: why is this not used?

packages/cli/src/task-runner.ts
```diff
@@ -140,6 +146,12 @@ export interface RunnerOptions {
 
 const logger = getLogger("TaskRunner");
 
+/**
+ * Maximum number of safe-to-batch tool calls to execute concurrently within a
+ * single microbatch in the CLI runner.
+ */
+const MaxToolCallConcurrency = 10;
```
zhanba: put in common, share with vscode

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+import {
+  checkReadOnlyConstraints,
+  isReadonlyToolCall,
+} from "./readonly-constraints-validation";
+
+export type ToolBatchMode = "concurrent" | "serial";
+
+export type ToolBatch<T> = {
+  mode: ToolBatchMode;
+  items: T[];
```
zhanba: for "serial" we only have one toolcall

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export async function runConcurrentBatch<T>(
+  items: T[],
+  concurrencyLimit: number,
+  execute: (item: T, batchMode: ToolBatchMode) => Promise<void>,
+): Promise<void> {
+  let nextIndex = 0;
+  const workerCount = Math.min(Math.max(concurrencyLimit, 1), items.length);
+  let firstError: unknown;
```
zhanba:
```suggestion

```

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+  | {
+      kind: "success";
+    }
+  | {
+      kind: "error";
+      error: string;
+    }
+  | {
+      kind: "cancelled";
+      reason: "user-abort" | "user-reject" | "previous-tool-call-failed";
```
zhanba: use QueueCancelReason

packages/vscode-webui/src/features/approval/components/tool-call-approval-button.tsx
```diff
@@ -214,7 +237,12 @@ export const ToolCallApprovalButton: React.FC<ToolCallApprovalButtonProps> = ({
     for (const lifecycle of lifecycles) {
```
zhanba: consider remove this?

packages/vscode-webui/src/features/chat/lib/chat-state/chat.tsx
```diff
@@ -14,6 +16,12 @@ export function ChatContextProvider({ children }: ChatContextProviderProps) {
   );
   const { executingToolCalls, getToolCallLifeCycle, completeToolCalls } =
     useToolCallLifeCycles(abortController.current.signal);
+  const { customAgents } = useCustomAgents(true);
+  const customAgentsRef = useRef(customAgents);
```
zhanba: uselatest?

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export type ScheduledToolCall = {
+  toolName: string;
+  input: unknown;
+  run: () => Promise<ScheduledToolCallResult>;
+  cancel: (reason: QueueCancelReason) => void;
+};
+
+/** Maximum concurrent executions within one concurrent batch. */
+const MaxConcurrency = 10;
```
zhanba: reuse

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  clearPending(reason: QueueCancelReason) {
+    const queue = this.queue.splice(0, this.queue.length);
+    this.cancelItems(queue, reason);
+  }
+
+  private async processAll(): Promise<void> {
+    try {
+      const queue = this.queue;
+      this.queue = [];
```
zhanba: clear after process toolcall

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+    this.processing = true;
+    this.processAll().catch((error) => {
+      if (!(error instanceof BatchExecutionError)) {
+        logger.error("Unexpected error in processAll", error);
+      }
+    });
+  }
+
+  clearPending(reason: QueueCancelReason) {
+    const queue = this.queue.splice(0, this.queue.length);
```
zhanba: rewrite in clear way

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (this.processing) return;
+    this.processing = true;
+    this.processAll().catch((error) => {
+      if (!(error instanceof BatchExecutionError)) {
+        logger.error("Unexpected error in processAll", error);
+      }
+    });
+  }
+
+  clearPending(reason: QueueCancelReason) {
```
zhanba:
```suggestion
  abort(reason: QueueCancelReason) {
```

packages/vscode-webui/src/features/chat/lib/batch-execute-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+        queue,
+        (item) => ({
+          toolName: item.toolName,
+          input: item.input,
+        }),
+        this.options.getCustomAgents?.(),
+      );
+
+      try {
+        await executePartitionedToolCalls(
```
zhanba: support abort controller for user cancel

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,221 @@
...
+      throw new BatchExecutionError(
+        BatchExecutionErrorMessages.FAILED,
+        error,
+        remainingItems,
+      );
+    }
+  }
+}
+
+export { checkReadOnlyConstraints, isReadonlyToolCall };
```
zhanba: extra export

packages/vscode-webui/src/features/tools/hooks/use-live-sub-task.tsx
```diff
@@ -18,37 +19,38 @@ import { useDefaultStore } from "@/lib/use-default-store";
...
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { createSubTaskToolCallAdapter } from "../../chat/lib/scheduled-tool-call-adapters";
 import type { ToolProps } from "../components/types";
 
 export function useLiveSubTask(
   { tool, isExecuting }: Pick<ToolProps<"newTask">, "tool" | "isExecuting">,
   toolCallStatusRegistry: ToolCallStatusRegistry,
 ): (TaskThreadSource & { parentId: string }) | undefined {
-  const lifecycle = useToolCallLifeCycle().getToolCallLifeCycle({
+  const { getToolCallLifeCycle } = useToolCallLifeCycle();
```
zhanba: extra change

packages/vscode-webui/src/features/tools/hooks/use-live-sub-task.tsx
```diff
@@ -18,37 +19,38 @@ import { useDefaultStore } from "@/lib/use-default-store";
...
     toolName: getStaticToolName(tool),
     toolCallId: tool.toolCallId,
   });
 
   const { customAgent, customAgentModel } = useCustomAgent(
     tool.state !== "input-streaming" ? tool.input?.agentType : undefined,
   );
+  // biome-ignore lint/style/noNonNullAssertion: uid must have been set.
+  const uid = tool.input?._meta?.uid!;
+  const subtaskCustomAgentsRef = useRef(
```
zhanba: use builtin agents

packages/vscode-webui/src/features/chat/lib/chat-state/fixed-state.tsx
```diff
@@ -118,6 +119,8 @@ export function FixedStateChatContextProvider({
 
   const completeToolCalls: FixedStateToolCallLifeCycle[] = [];
 
+  const batchExecuteManager = useRef(new BatchExecuteManager()).current;
```
zhanba: pass constructure param

packages/vscode-webui/src/features/tools/hooks/use-live-sub-task.tsx
```diff
@@ -151,76 +149,33 @@ export function useLiveSubTask(
...
+          abortSignal: abortController.current.signal,
           contentType: customAgentModel?.contentType,
           builtinSubAgentInfo,
           executeCommandWhitelist,
-          storeId: store.storeId,
-          taskId: uid,
+          addToolOutput,
+          toolCallStatusRegistry,
+        }),
+        {
```
zhanba: no need

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,227 @@
...
+  | {
+      isConcurrencySafe: true;
+      items: ScheduledToolCall[];
+    }
+  | {
+      isConcurrencySafe: false;
+      items: [ScheduledToolCall];
+    };
+
+export type QueueCancelReason = "user-abort" | "previous-tool-call-failed";
```
zhanba:
```suggestion
export type ToolCallCancelReason = "user-abort" | "previous-tool-call-failed";
```

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,227 @@
...
+    batches.push({
+      isConcurrencySafe: true,
+      items: currentConcurrentBatch,
+    });
+  }
+
+  return batches;
+}
+
+export async function runConcurrentBatch(
```
zhanba, line 139:
```suggestion
async function runConcurrentBatch(
```

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,227 @@
...
+ * Partition an ordered list of tool-call-backed items into microbatches:
+ *
+ * - Consecutive safe-to-batch calls → one concurrent batch.
+ * - Each stateful call → its own single-element serial batch.
+ *
+ * Batches execute sequentially; batch N+1 only starts after N fully completes.
+ */
+export function partitionToolCalls(
+  items: ScheduledToolCall[],
+  getToolCall: (item: ScheduledToolCall) => {
```
zhanba: seems useless

packages/tools/src/utils/batch-utils.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Partition an ordered list of tool-call-backed items into microbatches:
+ *
+ * - Consecutive safe-to-batch calls → one concurrent batch.
+ * - Each stateful call → its own single-element serial batch.
+ *
+ * Batches execute sequentially; batch N+1 only starts after N fully completes.
+ */
+export function partitionToolCalls(
```
zhanba, line 101: partitionToolCalls and executePartitionedToolCalls can combine into one function as they always use together; we can just use partitionToolCalls inside executePartitionedToolCalls

packages/vscode-webui/src/features/chat/lib/scheduled-tool-call-adapters.ts
```diff
@@ -0,0 +1,283 @@
...
+  abortSignal: AbortSignal;
+  contentType?: string[];
+  builtinSubAgentInfo?: BuiltinSubAgentInfo;
+  executeCommandWhitelist?: string[];
+  addToolOutput: ReturnType<typeof useChat>["addToolOutput"];
+  toolCallStatusRegistry: ToolCallStatusRegistry;
+};
+
+/** Backed by a ToolCallLifeCycle; execution and cancellation delegate to the lifecycle. */
+export function createLifecycleToolCallAdapter({
```
zhanba:
```suggestion
export function createBatchedToolCallFromLifecycle({
```

packages/vscode-webui/src/features/chat/lib/scheduled-tool-call-adapters.ts
```diff
@@ -0,0 +1,283 @@
...
+      });
+    },
+    cancel: (reason: QueueCancelReason) => {
+      lifecycle.abort(reason);
+    },
+  };
+}
+
+/** For offhand-mode sub-tasks: directly executes via vscodeHost and reports results with addToolOutput. */
+export function createSubTaskToolCallAdapter({
```
zhanba:
```suggestion
export function createSubtaskBatchedToolCall({
```

PR #1500, opened 2026-04-28
feat(tools): implement domain-based tool policies for webSearch

liangfung:
## Summary
- Implemented domain-based tool policies for the `webSearch` tool, similar to `webFetch`.
- Refactored domain policy compilation and validation to be generic across `webFetch` and `webSearch`.
- Enhanced `webSearch` validation to automatically inject agent-configured allowed domains into `searchDomainFilter` if missing or only containing denylist entries.
- Added comprehensive tests for `webSearch` domain policy enforcement and automatic allowlist injection.

## Test plan
- Run `bun vitest packages/tools/src/__test__/utils.test.ts` to verify all domain policy tests pass.
- Verified that `searchDomainFilter` is correctly mutated when missing or deny-only to include agent-configured allowlist.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-361eab69307044f587baa6603b34f09e)

### DIFF
--- a/packages/tools/src/__test__/utils.test.ts
+++ b/packages/tools/src/__test__/utils.test.ts
@@ -300,3 +300,81 @@ describe("webFetch domain policies", () => {
     ).toThrow("URL domain is not allowed by the configured webFetch domain rules.");
   });
 });
+
+describe("webSearch domain policies", () => {
+  it("should compile domain pattern rules for webSearch", () => {
+    const policies = compileToolPolicies([
+      "webSearch(domain:example.com)",
+      "webSearch(domain:*.tabbyml.com)",
+    ]);
+
+    expect(policies?.webSearch).toEqual({
+      kind: "domain-pattern",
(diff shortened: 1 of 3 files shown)


liangfung:
`webSearch.input.searchDomainFilter` supports either allow or deny mode, while our agent-configured domain rules are allowlist-based. This makes the behavior unclear when combining the two sources of rules.

More specifically:
- If `searchDomainFilter` is absent, applying the configured allowlist is straightforward.
- If `searchDomainFilter` is in allow mode, we would need to intersect it with the configured allowlist.
- If `searchDomainFilter` is in deny mode, there is no natural way to merge it with an allowlist without changing the intended semantics.

AI-provided filter rules may also differ from our configured rules, so merging them would be brittle.

Because of this mismatch between allow/deny filters and allowlist-based configuration, we are closing this PR for now.

PR #1665, opened 2026-06-16
fix(vscode-webui): refine checkpoint layout spacing

godtail:
## Summary
- Dynamically calculate the top margin of a message part (`paddingClass`). If the previous part is a `data-checkpoint`, do not add the top margin (`mt-2`). This prevents unnecessary layout spacing/padding between a checkpoint and the following text or content.

## Test plan
- Verify that checkpoints align perfectly in the message list without layout shifts or extra spacing.
- Run `bun check` and `bun tsc` to ensure all checks pass.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-849aa791624341b695436e4f91f8e25f)

### DIFF
--- a/packages/vscode-webui/src/components/message/message-list.tsx
+++ b/packages/vscode-webui/src/components/message/message-list.tsx
@@ -173,6 +173,7 @@ export const MessageList: React.FC<{
                       }
                       partIndex={index}
                       part={part}
+                      previousPart={m.parts[index - 1]}
                       isLoading={isLoading}
                       isExecuting={isExecuting}
                       messages={renderMessages}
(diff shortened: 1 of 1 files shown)


packages/vscode-webui/src/components/message/message-list.tsx
```diff
@@ -173,6 +173,7 @@ export const MessageList: React.FC<{
                       }
                       partIndex={index}
                       part={part}
+                      previousPart={m.parts[index - 1]}
```
zhanba, line 176: this seems not very robust for first part

godtail: m.parts[-1] will return undefined here when index is 0

PR #1668, opened 2026-06-17
fix(vscode-webui): improve checkpoint ui layout and spacing

godtail:
## Summary
- Reverts absolute vertical centering layout changes for the checkpoint button introduced in #1616.
- Positions the checkpoint button with absolute top positioning and adds appropriate margin wrappers to prevent layout overlapping.
- Removes `content-visibility` optimizations from `.message-list-item` to prevent layout shifts and measurement loops on virtualized scroll.

image

## Test plan
- Run webui unit tests: `bun vitest --run` inside `packages/vscode-webui` to verify layout components.
- Manually verify checkpoint buttons are correctly positioned and aligned in the chat message list without overlapping adjacent text.
- Scroll through long chat lists to ensure no layout jitter or rendering loops.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-147962ab9078431a9052821ea5536805)

### DIFF
--- a/packages/vscode-webui/src/styles.css
+++ b/packages/vscode-webui/src/styles.css
@@ -25,14 +25,6 @@ code {
     monospace;
 }
 
-.message-list-item {
-  contain-intrinsic-size: auto 240px;
-}
-
-.perf-content-visibility-off .message-list-item {
-  contain-intrinsic-size: none;
-}
-
 :root {
   --background: oklch(1 0 0);
   --foreground: oklch(0.141 0.005 285.823);
--- a/packages/vscode-webui/src/styles.css
+++ b/packages/vscode-webui/src/styles.css
@@ -26,12 +26,10 @@ code {
 }
 
 .message-list-item {
-  content-visibility: auto;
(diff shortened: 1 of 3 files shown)


packages/vscode-webui/src/styles.css
```diff
@@ -26,12 +26,10 @@ code {
 }
 
 .message-list-item {
```
liangfung: remove the entire class name

packages/vscode-webui/src/styles.css
```diff
@@ -26,12 +26,10 @@ code {
...
-  content-visibility: auto;
   contain-intrinsic-size: auto 240px;
 }
 
 .perf-content-visibility-off .message-list-item {
```
liangfung: remove the entire class name

packages/vscode-webui/src/styles.css
godtail: Also remove `contain-intrinsic-size`?

PR #1759, opened 2026-07-06
feat(vscode-webui): support steering queued messages

godtail:
## Summary
- Introduce a steering mechanism for queued messages in the chat interface.
- Redesign the queued messages display into a compact layout with dedicated "Steer" and "Remove" actions.
- Add comprehensive unit tests for the chat submit hook.

e77ccc497966cf066c7790b6c00ac4f1

## Test plan
- Run the added unit tests: `bun run test packages/vscode-webui/src/features/chat/hooks/use-chat-submit.test.tsx`
- Manually queue messages and verify they can be steered (stopped and promoted to the front) or removed.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-4a308944df4e4c12b1a937e4090ee735)


packages/vscode-webui/src/features/chat/hooks/use-chat-submit.ts
```diff
@@ -161,13 +237,13 @@ export function useChatSubmit({
           // Error is already handled by the hook
           return;
         }
-      } else if (allMessages.length > 0 || reviews.length > 0) {
+      } else if (text.length > 0 || reviews.length > 0) {
```
liangfung: Queued review-only messages can be dropped here. When flushing from the queue, this condition checks the current `reviews.length`, but the queued reviews are in `messageReviews`. Since current reviews are cleared after queuing, `{ text: "", files: [], reviews: [...] }` passes validation but never gets sent.

liangfung:
```suggestion
      } else if (text.length > 0 || messageReviews.length > 0) {
```

PR #1785, opened 2026-07-13
fix: various Windows compatibility issues

zhanba:
## Summary

This PR bundles several Windows-specific fixes.

### 1. Command output encoding (mojibake / 乱码)
- On Chinese Windows, PowerShell/cmd emit command output using the console's OEM codepage (GBK/CP936), but Pochi decoded the bytes as UTF-8, producing mojibake — e.g. PowerShell's localized "git 无法将..." error rendered as garbage.
- Force the Windows shells to emit UTF-8 at the source in `buildShellCommand`: PowerShell gets a BOM-less `[Console]::OutputEncoding=[System.Text.UTF8Encoding]::new($false);` prefix; cmd gets a `chcp 65001>nul &` prefix.
- Decode child-process stdout/stderr with `node:string_decoder`'s `StringDecoder` (flushed on close) so multi-byte characters split across data chunks are no longer corrupted.
- This affects the `executeCommandWithNode` path, the only active command path on Windows (node-pty is disabled there).

### 2. Path comparison mismatches
- git emits forward-slash, upper-cased-drive paths (`C:/Users/foo/repo`) while VS Code's `Uri.fsPath` emits back-slash, lower-cased-drive paths (`c:\Users\foo\repo`). These never matched, which hid tasks/worktrees from the task list on Windows.
- Added `normalizePathForComparison` (unify separators, lower-case drive letter, strip trailing slashes) and use it wherever cwd/worktree/gitdir paths are compared (task list, worktree list, git-status reader, vscode-host worktree scope).

### 3. ripgrep execution & binary resolution
- Run ripgrep via `execFile` with an argument array instead of a shell command string. Manual single-quote quoting broke on Windows because cmd.exe does not strip single quotes, so rg received literal quotes around the path and failed with "path not found".
- Probe `node_modules.asar.unpacked` (in addition to `node_modules`) when resolving VS Code's bundled `rg` binary, since native binaries are extracted alongside the asar archive.

## Test plan
- [x] `packages/common` shell tests updated to assert the new UTF-8 shell prefixes
- [x] Added a split-multi-byte decode test in `execute-command-with-node.test.ts`
- [x] Added `normalizePathForComparison` unit tests
- [x] Updated ripgrep tests to use `execFile` + platform-appropriate paths
- [x] `bun run test` passes for both packages
- [x] Biome check clean

> Note: the CLI (`packages/cli`) has the same class of encoding bug in its own exec/spawn paths, but the reported issue is in the VS Code extension so it's out of scope here.

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-4bdfd78ecca0433faba75d8927471fe7)


zhanba (commented):
.

PR #1812, opened 2026-07-20
feat: support terminal text selection as active-selection context

icycodes:
## Summary
- Add a sibling `activeTerminalTextSelection` field to the `data-active-selection` part (kept independent from the existing `ActiveSelection` schema to avoid any migration risk).
- Add `readTerminalSelection` host API that snapshots/restores the clipboard around `workbench.action.terminal.copySelection` to read the terminal's current selection, since VS Code has no stable API for this (microsoft/vscode#188173).
- Wire the new field through message preparation, chat submit, and new-task creation flows.
- Render a terminal chip (`TerminalSelectionPart`) in the message list with a hover preview and click-to-jump-to-terminal behavior.
- Render both selection kinds as separate text blocks when sending context to the LLM.

## Test plan
- [ ] `bun run test` passes for affected packages (common, livekit, vscode-webui, vscode)
- [ ] Manually select text in an integrated terminal, send a chat message, and verify the terminal selection chip appears and its content is forwarded to the LLM
- [ ] Verify existing file-based active-selection behavior is unaffected

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-5139e8bfb20c45ccb6736856fc338405)


packages/common/src/vscode-webui-bridge/types/message.ts
```diff
@@ -49,6 +49,21 @@ export const ActiveSelection = z
 
 export type ActiveSelection = z.infer<typeof ActiveSelection>;
 
+export const TerminalTextSelection = z.object({
```
zhanba: This file is a duplicate of the previous one

icycodes: Updated

PR #1867, opened 2026-08-03
feat: support content-filtering metadata extraction and retry handling

liangfung:
## Summary
* Extracted content filter metadata from chat completions when the finish reason is `content-filter` or when Anthropic/Google content blocks are present.
* Disabled automatic retries for content-filtered errors and displayed localized safety filter warnings to the user.
* Updated UI components (like the retry approval button, error message view, and live subtasks) to support explicit manual retry instead of "Continue" or automatic retries.

## Screenshot
image

## Test plan
- [x] Verified unit tests in `packages/livekit` (`flexible-chat-transport.test.ts`) using `bun test` to ensure content filter metadata is correctly extracted.
- [x] Verified unit tests in `packages/vscode-webui` (including `use-retry.test.tsx` and `use-live-sub-task.test.tsx`) under Node-based `npx vitest` run, and all 12/12 unit tests pass successfully.

### DIFF
--- a/packages/vscode-webui/src/features/retry/hooks/use-retry.ts
+++ b/packages/vscode-webui/src/features/retry/hooks/use-retry.ts
@@ -30,15 +30,6 @@ export function useRetry({
 
       if (
         error instanceof ReadyForRetryError &&
-        error.kind === "content-filter"
-      ) {
-        return regenerate({
-          messageId: lastMessage.id,
-        });
-      }
-
-      if (
-        error instanceof ReadyForRetryError &&
         error.kind === "no-tool-calls"
       ) {
         return sendMessage({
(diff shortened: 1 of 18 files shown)


packages/vscode-webui/src/features/retry/hooks/use-retry.ts
```diff
@@ -28,6 +28,15 @@ export function useRetry({
         return sendMessage(undefined);
       }
 
+      if (
+        error instanceof ReadyForRetryError &&
+        error.kind === "content-filter"
+      ) {
+        return regenerate({
```
zhanba: need consider clearFileStateCache if we reset message that have file releated toolcall

PR #1882, opened 2026-08-07
docs: add isolated-workspaces video to weekly update #36

jinsyl123:
Adds the isolated-workspaces demo video to the already-published Weekly Update #36 (the entry was missing its #1820 media). Video hosted on R2. No other content changes.


jinsyl123:
weekly update #36 is dated 07-30 but the bi-weekly cadence would put it at 08-07, was that intentional? 

image

PR #1924, opened 2026-08-31
fix(livekit): bound memory background task steps

zhanba:
## Summary
- allow forked background tasks to persist a task-specific maximum step count
- cap task memory at 3 steps, auto memory extraction at 5, and auto memory dream consolidation at 20
- retain the existing 50-step fallback for background tasks without an explicit limit

## Test plan
- [x] Run targeted LiveKit Vitest suites (22 tests)
- [x] Run `bun check`
- [x] Run `bun tsc`
- [x] Run `git diff --check`

🤖 Generated with [Pochi](https://getpochi.com) | [Task](https://app.getpochi.com/share/p-756b25d05481454296b52ae4fcf65ad7)

### DIFF
--- a/packages/livekit/src/background-task/memory/task-memory.ts
+++ b/packages/livekit/src/background-task/memory/task-memory.ts
@@ -43,6 +43,7 @@ const TaskMemoryAllowedTools: readonly ToolSpecInput[] = [
   "readFile",
   `writeToFile(${TaskMemoryFileUri})`,
 ];
+const TaskMemoryMaxSteps = 3;
 
 const TaskMemoryStoreFilePath = new URL(TaskMemoryFileUri).pathname;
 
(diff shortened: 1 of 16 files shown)


packages/livekit/src/background-task/memory/task-memory.ts
```diff
@@ -53,27 +51,32 @@ function getExtractionMetrics<TMessage extends UIMessage>(data: {
...
-
-  const tokenDelta = metrics.tokens - state.lastExtractionTokens;
-  const toolCallDelta = metrics.toolCalls - state.lastExtractionToolCalls;
+  if (metrics.tokens < trigger) return false;
+  if (state.extractedSinceCompact) return false;
 
   return (
-    tokenDelta >= constants.TaskMemoryUpdateTokenIncrement &&
-    toolCallDelta >= constants.TaskMemoryUpdateToolCallThreshold
+    state.extractionAttemptsSinceCompact <
```
zhanba, line 78: [P2] Normalize persisted task-memory state before reading this field. Tasks saved by the previous release have `initialized`/`lastExtractionTokens` but no `extractionAttemptsSinceCompact`; for those tasks this comparison becomes `undefined < 2` (false), and the increment below would produce `NaN`, so task-memory extraction is silently disabled for the whole first post-upgrade compaction cycle. Merge the stored value with `DefaultTaskMemoryState` (or migrate it) when loading.

PR #1931, opened 2026-09-02
fix(webhooks): include subtasks in delivery scan

### LINKED ISSUE #1663: Bug: parent id is not included in livekit-cf webhook
**Describe the bug**

Parent id is not included in livekit-cf webhook, making all the parent id is null in ragdoll.

**Information about your version**
Please provide version of Pochi.

**Additional context**

pengpengyi92:
## Summary

- keep the existing root-task query used by task-list consumers
- add an all-task query for webhook delivery
- include updated subtasks in the Cloudflare webhook scan so their real
  `parentId` reaches the existing payload serializer
- add a regression that pins root-only and all-task query semantics separately

Fixes #1663

## Root cause

The webhook path reused `tasks$`, which intentionally filters
`parentId = null`. Child rows were therefore excluded before payload
serialization; changing the serializer alone could not repair the behavior.

## Test plan

- LiveKit package test suite (`142 passed`)
- TypeScript checks for `packages/livekit` and `packages/livekit-cf`
- Biome check for all modified files
- `git diff --check`

## Boundary

No database schema or webhook payload contract changes. The patch only changes
which recently updated task rows are eligible for webhook delivery. The local
pre-push script also requires Bun, which is unavailable on this workstation;
the focused suites above passed and the upstream CI remains the full-repository
gate.

### DIFF
--- a/packages/livekit-cf/src/do/livestore-client/do.ts
+++ b/packages/livekit-cf/src/do/livestore-client/do.ts
@@ -105,7 +105,7 @@ export class LiveStoreClientDO
     await this.getStore();
     // // Make sure to only subscribe once
     // if (this.storeSubscription === undefined) {
-    //   this.storeSubscription = store.subscribe(catalog.queries.tasks$, {
+    //   this.storeSubscription = store.subscribe(catalog.queries.allTasks$, {
     //     // FIXME(meng): implement this with store.events stream when it's ready
     //     onUpdate: (tasks) => this.onTasksUpdateThrottled.call(tasks),
     //   });
@@ -126,7 +126,7 @@ export class LiveStoreClientDO
 
   private onTasksUpdate = async (force?: boolean) => {
     const store = await this.getStore();
-    const tasks = store.query(catalog.queries.tasks$);
+    const tasks = store.query(catalog.queries.allTasks$);
     const oneMinuteAgo = moment().subtract(1, "minute");
 
     const updatedTasks = tasks.filter(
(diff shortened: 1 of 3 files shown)


zhanba:
Thanks for the contribution! Could you share the real-world use case behind this change?

pengpengyi92:
Thanks for checking. I do not operate the reporter's Ragdoll deployment, so I
cannot claim an independently captured production payload. I took the concrete
use case from #1663 and verified the corresponding source path:

1. `v1.TaskInited` persists a child task's `parentId` in `tables.tasks`.
2. `LiveStoreClientDO.onTasksUpdate()` used `tasks$`, whose explicit
   `parentId = null` predicate returns only root tasks.
3. `WebhookDelivery` spreads the task row into the payload, but child rows
   never reached that serializer. A downstream consumer such as the Ragdoll
   integration described in #1663 therefore receives root-task updates only
   and cannot reconstruct the child-to-parent relationship.

The patch does not synthesize or rewrite `parentId`; it lets an updated child
row reach the existing `task.updated` payload with its stored parent ID. The
trade-off is that webhook volume now includes updated subtasks as well as root
tasks.

If the intended webhook contract is root tasks only, this PR should be closed.
If integrations are expected to consume subtask lifecycle events, I can replace
the current query-contract regression with a focused delivery-path test before
merge. A sanitized payload or expected Ragdoll contract from the reporter would
also let me validate that boundary more directly.

